PR #5, opened 2022-10-27
DPE-835: Dynamic generation of spark-defaults.conf. Initial design for discussion.

averma-canonical:
Changes in this PR are currently only the proposal of how we want to structure DPE-835 solution. Once we concur, I can materialize the implementation and start investing in testing.

Idea:
spark-defaults.conf needs to be dynamically generated and consulted in spark-submit before applying per submit command conf overrides.

The **defaults config file** is to be in order of decreasing priority
- $SNAP_SPARK_ENV_CONF 
- $SPARK_CONF_DIR/conf/spark-defaults.conf
- $SPARK_HOME/conf/spark-defaults.conf
- $SNAP/conf/spark-defaults.conf

1. Default set of configs are dynamically generated including the username and namespace specified in setup command.
2. config from the **defaults config file** is read.
3. Values in [2] override the ones already present in [1]
4. Composite config is now written back to the same **defaults config file** i.e. an **overwrite**/augmentation happens

### DIFF
--- a/helpers/setup-spark-k8s.py
+++ b/helpers/setup-spark-k8s.py
@@ -6,6 +6,7 @@ import yaml
 import argparse
 import pwd
 from typing import Dict
+import utils
 
 KUBECTL_CMD= '{}/kubectl'.format(os.environ['SNAP'])
 USER_HOME_DIR_ENT_IDX = 5
@@ -104,6 +105,19 @@ def set_up_user(username: str, name_space: str, defaults: Dict) -> None:
     os.system(f"{KUBECTL_CMD} label serviceaccount --kubeconfig={kubeconfig} --context={context_name} {username} {label} --namespace={namespace}")
     os.system(f"{KUBECTL_CMD} label rolebinding --kubeconfig={kubeconfig} --context={context_name} {rolebindingname} {label} --namespace={namespace}")
 
+def setup_spark_conf_defaults(username: str, namespace: str) -> None:
+    SPARK_CONF_DEFAULTS_FILE = utils.get_spark_defaults_conf_file()
+    generated_defaults = utils.generate_spark_default_conf()
+    if username:
+        generated_defaults['spark.kubernetes.authenticate.driver.serviceAccountName'] = username
+
+    if namespace:
+        generated_defaults['spark.kubernetes.namespace'] = namespace
+
+    user_provided_defaults = utils.read_spark_defaults_file(SPARK_CONF_DEFAULTS_FILE)
+    final_defaults = utils.override_conf_defaults(generated_defaults, user_provided_defaults)
+    utils.write_spark_defaults_file(SPARK_CONF_DEFAULTS_FILE, final_defaults)
+
 if __name__ == "__main__":
     USER_HOME_DIR = pwd.getpwuid(os.getuid())[USER_HOME_DIR_ENT_IDX]
     DEFAULT_KUBECONFIG = f'{USER_HOME_DIR}/.kube/config'
@@ -127,3 +141,4 @@ if __name__ == "__main__":
         username = args.username
         namespace = args.namespace or defaults['namespace']
         set_up_user(username, namespace, defaults)
+        setup_spark_conf_defaults(username, namespace)
--- a/helpers/spark-submit.py
+++ b/helpers/spark-submit.py
@@ -6,6 +6,7 @@ import pwd
 import subprocess
 import logging
 import argparse
+import utils
 
 USER_HOME_DIR_ENT_IDX = 5
 EXIT_CODE_BAD_KUBECONFIG = -100
@@ -29,18 +30,25 @@ if __name__ == "__main__":
         os.environ['SPARK_HOME'] = os.environ['SNAP']
 
     submit_args = sys.argv[1:]
-    conf = dict()
 
+    SPARK_CONF_DEFAULTS_FILE = utils.get_spark_defaults_conf_file()
+    conf_defaults = utils.read_spark_defaults_file(SPARK_CONF_DEFAULTS_FILE)
+    conf_overrides = dict()
     if args.conf:
         for c in args.conf:
             try:
                 kv = c.split('=')
                 k = kv[0]
                 v = '='.join(kv[1:])
-                conf[k] = os.environ.get(v, v)
+                conf_overrides[k] = os.environ.get(v, v)
             except IndexError as e:
                 logging.error('Configuration related arguments parsing error. Please check input arguments and try again.')
                 sys.exit(EXIT_CODE_BAD_CONF_ARG)
+
+    conf = utils.override_conf_defaults(conf_defaults, conf_overrides)
+
+    # TBD: Needs discussion
+    # submit_args = utils.reconstruct_submit_args(args, conf)
 
     if not args.master:
         USER_HOME_DIR = pwd.getpwuid(os.getuid())[USER_HOME_DIR_ENT_IDX]
--- a/helpers/utils.py
+++ b/helpers/utils.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python3
+
+import os
+from typing import List, Dict
+import re
+import pwd
+import argparse
+
+USER_HOME_DIR_ENT_IDX = 5
+
+def generate_spark_default_conf() -> Dict:
+    defaults = dict()
+    defaults['spark.app.name'] = 'spark-app'
+    defaults['spark.executor.instances'] = 2
+    defaults['spark.kubernetes.container.image'] = 'docker.io/averma32/sparkpy6:latest'
+    defaults['spark.kubernetes.container.image.pullPolicy'] = 'IfNotPresent'
+    defaults['spark.kubernetes.namespace'] = 'default'
+    defaults['spark.kubernetes.authenticate.driver.serviceAccountName'] = 'spark'
+    defaults['spark.eventLog.enabled'] = 'false'
+    USER_HOME_DIR = pwd.getpwuid(os.getuid())[USER_HOME_DIR_ENT_IDX]
+    SCALA_HIST_FILE_DIR = os.environ.get('SNAP_USER_COMMON', USER_HOME_DIR)
+    defaults['spark.driver.extraJavaOptions'] = f'-Dscala.shell.histfile={SCALA_HIST_FILE_DIR}/.scala_history'
+
+    return defaults
+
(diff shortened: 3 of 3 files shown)


deusebio (commented):
The code looks very fine and I like how this is shaping up, with the user experience. 

I only have one issue to be addressed before merging. At the moment, all configurations from SNAP, ServiceAccount, ENV, extra, etc are all merged into a dictionary that it is then provided to submit via --conf. If we have many many configurations, this may not be scalable and it is probably best to write down a file, and provide that with "--properties-file"

Beside, we are already writing out the file, but we are actually not using it at the moment. What I would do is to write the file down and provided that with properties flag. I would only leave out of the file, the "--conf" directly provided by the user

helpers/utils.py
```diff
@@ -0,0 +1,102 @@
...
+import logging
+import io
+
+USER_HOME_DIR_ENT_IDX = 5
+EXIT_CODE_BAD_KUBECONFIG = -100
+EXIT_CODE_BAD_CONF_ARG = -200
+
+def generate_spark_default_conf() -> Dict:
+    generated_defaults = dict()
+    generated_defaults['spark.kubernetes.container.image'] = 'docker.io/averma32/sparkpy6:latest'
```
deusebio, line 18: This is not to be overwritten

helpers/utils.py
```diff
@@ -0,0 +1,102 @@
...
+    generated_defaults['spark.kubernetes.container.image.pullPolicy'] = 'IfNotPresent'
```
deusebio, line 19: I would also not overwrite this configurations

helpers/spark-submit.py
```diff
@@ -16,55 +16,35 @@
...
 
-    if not args.master:
-        USER_HOME_DIR = pwd.getpwuid(os.getuid())[USER_HOME_DIR_ENT_IDX]
-        DEFAULT_KUBECONFIG = f'{USER_HOME_DIR}/.kube/config'
-        kubeconfig = os.environ.get('KUBECONFIG') or DEFAULT_KUBECONFIG
-        kubectl_cmd = '{}/kubectl'.format(os.environ['SNAP'])
+    static_defaults = utils.read_property_file(STATIC_DEFAULTS_CONF_FILE)
+    dynamic_defaults = utils.read_property_file(DYNAMIC_DEFAULTS_CONF_FILE) if os.path.isfile(DYNAMIC_DEFAULTS_CONF_FILE) else dict()
+    user_defaults = utils.read_property_file(args.properties_file) if args.properties_file else dict()
+    user_overrides = utils.parse_conf_overrides(args.conf) if args.conf else dict()
```
deusebio, line 34: My idea was to provide the "--conf" flags straight into the submit command, with no parsing/creationg of the dict. But it is not a strong point

helpers/utils.py
```diff
@@ -0,0 +1,102 @@
...
+def get_snap_defaults_conf_file() -> str:
+    SPARK_HOME = os.environ.get('SPARK_HOME', os.environ.get('SNAP'))
+    SPARK_CONF_DIR = os.environ.get('SPARK_CONF_DIR', f"{SPARK_HOME}/conf")
+    SPARK_CONF_DEFAULTS_FILE = os.environ.get('SNAP_SPARK_ENV_CONF', f"{SPARK_CONF_DIR}/spark-defaults.conf")
+    return SPARK_CONF_DEFAULTS_FILE
+
+def get_static_defaults_conf_file() -> str:
+    return os.environ.get('SNAP') + '/conf/spark-defaults.conf'
+
+def get_dynamic_defaults_conf_file() -> str:
```
deusebio, line 55: I would call this "get_env_defaults_conf_file". Dynamic is somewhat confusing since the dynamic part will be the task DPE-836, and I would just to stick to the naming we had in the slides

helpers/setup-spark-k8s.py
```diff
@@ -104,6 +105,17 @@ def set_up_user(username: str, name_space: str, defaults: Dict) -> None:
     os.system(f"{KUBECTL_CMD} label serviceaccount --kubeconfig={kubeconfig} --context={context_name} {username} {label} --namespace={namespace}")
     os.system(f"{KUBECTL_CMD} label rolebinding --kubeconfig={kubeconfig} --context={context_name} {rolebindingname} {label} --namespace={namespace}")
 
+def setup_spark_conf_defaults(username: str, namespace: str) -> None:
+    DYNAMIC_DEFAULTS_CONF_FILE = utils.get_dynamic_defaults_conf_file()
```
deusebio, line 109: I would also use an env variable to specify this, as we discussed in the slides. This should allow us to specify different environment, between prod, test, svil environments (that might have different resource demands...e.g. executors)

helpers/spark-submit.py
```diff
@@ -16,55 +15,32 @@
...
+    with utils.UmaskNamedTemporaryFile(mode = 'w', prefix='spark-conf-', suffix='.conf') as t:
+        defaults = utils.merge_configurations([snap_static_defaults, setup_dynamic_defaults, env_defaults, props_file_arg_defaults])
+        logging.debug(f'Spark props available for reference at {utils.get_snap_temp_dir()}{t.name}\n')
+        utils.write_property_file(t.file, defaults, log=True)
+        t.flush()
+        submit_args = [f'--master {args.master or utils.autodetect_kubernetes_master(defaults)}',
+                       f'--deploy-mode {args.deploy_mode}',
+                       f'--properties-file {t.name}'] + extra_args
+        submit_cmd = '{}/bin/spark-submit'.format(os.environ['SPARK_HOME'])
+        submit_cmd += ' ' + ' '.join(submit_args)
```
deusebio, line 44: why not using f-string interpolation in one go?

helpers/spark-submit.py
```diff
@@ -16,55 +15,32 @@
...
-
-    submit_cmd += ' ' + ' '.join(submit_args)
-    os.system(submit_cmd)
\ No newline at end of file
+    STATIC_DEFAULTS_CONF_FILE = utils.get_static_defaults_conf_file()
+    DYNAMIC_DEFAULTS_CONF_FILE = utils.get_dynamic_defaults_conf_file()
+    ENV_DEFAULTS_CONF_FILE = utils.get_env_defaults_conf_file()
+
+    snap_static_defaults = utils.read_property_file(STATIC_DEFAULTS_CONF_FILE)
+    setup_dynamic_defaults = utils.read_property_file(DYNAMIC_DEFAULTS_CONF_FILE)
```
deusebio, line 31: I would probably put the "if file exists" statement here too. Just to be safe.

deusebio (approved):
Looks good! Only some minor comments attached. 

I'd open this for review to the other reviewers

helpers/spark-defaults.conf
```diff
@@ -1,8 +1,2 @@
-spark.app.name=spark-app
-spark.executor.instances=2
 spark.kubernetes.container.image=docker.io/averma32/sparkpy6:latest
-spark.kubernetes.container.image.pullPolicy=IfNotPresent
-spark.kubernetes.namespace=default
-spark.kubernetes.authenticate.driver.serviceAccountName=spark
-spark.eventLog.enabled=false
-spark.driver.extraJavaOptions='-Dscala.shell.histfile=${SNAP_USER_COMMON}/.scala_history'
\ No newline at end of file
+spark.kubernetes.container.image.pullPolicy=Always
```
grobbie, line 2: Is a pullpolicy required? If not could we remove it?

helpers/utils.py
```diff
@@ -0,0 +1,111 @@
...
+       logging.error(e.output)
+       sys.exit(EXIT_CODE_BAD_KUBECONFIG)
+    return k8s_master_uri
+
+def UmaskNamedTemporaryFile(*args, **kargs):
+    fdesc = NamedTemporaryFile(*args, **kargs)
+    umask = os.umask(0o666)
+    os.umask(umask)
+    os.chmod(fdesc.name, 0o666 & ~umask)
+    return fdesc
```
grobbie, line 111: Please can we add some unit tests for the script and some functional tests that demonstrate everything works, thanks

helpers/spark-submit.py
```diff
@@ -16,55 +15,34 @@
...
+        t.flush()
+        submit_args = [f'--master {args.master or utils.autodetect_kubernetes_master(defaults)}',
+                       f'--deploy-mode {args.deploy_mode}',
+                       f'--properties-file {t.name}'] + extra_args
+
+        SPARK_HOME = os.environ['SPARK_HOME']
+        SPARK_SUBMIT_ARGS = ' '.join(submit_args)
+        submit_cmd = f'{SPARK_HOME}/bin/spark-submit {SPARK_SUBMIT_ARGS}'
+        logging.debug(submit_cmd)
+        os.system(submit_cmd)
```
grobbie, line 48: Needs some tests

grobbie (changes requested):
Generally this is starting to shape up really well. Needs some tests to round it out. Great work so far.

deusebio:
> Needs some tests to round it out.

Yes, @grobbie ! I totally agree! As a matter of fact, we have been talking about this in the past few days with @averma-canonical . We have already placed an item in the next pulse for addressing the design/implementation of a skeleton for the tests (DPE-854) and then single tests implementation (DPE-855).

.github/workflows/snap.yaml
```diff
@@ -27,27 +25,27 @@ jobs:
...
-        name: Read last published stable version
+        name: Read last published stable Spark version
         uses: juliangruber/read-file-action@v1
         with:
-          path: ./VERSION
+          path: ./SPARK_VERSION
       - id: latest-version
         name: Check latest stable version available for download
         run: |
           LATEST_STABLE_SPARK_VERSION=$(curl --silent https://downloads.apache.org/spark/ | grep "spark-" | cut -d'>' -f3 | cut -d'/' -f1  | sort | tail -n 1)
```
welpaolo, line 37: Can you please make two different variables one for the full version name `spark-3.3.1` and another one with the version only `3.3.1` ?

welpaolo (dismissed):
Just a minor comment.

deusebio (dismissed):
I have talked to @grobbie . We agreed that we can go ahead with merging this PR but within the next sprint we **MUST** fix the framework for testing, and allow PR merge ONLY IF testing is done!

welpaolo (approved):
LGTM

deusebio (approved):
Good to go!

PR #6, opened 2022-11-07
DPE-838: Enabling Spark service account configuration setup as k8s secrets

averma-canonical:
We want to provide a way for the users to set configurations at the service account level, stored in kubernetes as secrets named “spark-client-sa-conf-${SERVICE_ACCOUNT}” with data as 

{“spark.executor.instances”: …}

We want to have some API to create or view properties 
setup_spark_k8s sa-conf-create --service-account-name SERVICE_ACCOUNT_NAME --properties-file PROPERTIES_FILE 
setup_spark_k8s sa-conf-get --service-account-name SERVICE_ACCOUNT_NAME [--conf]

Coming Soon:
Remove properties (except for the namespace and serviceaccount name)

### DIFF
--- a/helpers/setup-spark-k8s.py
+++ b/helpers/setup-spark-k8s.py
@@ -141,6 +141,10 @@ if __name__ == "__main__":
                                 help='Service Account for which configuration properties need to be retrieved.')
     parser_account.add_argument('--conf', action='append', type=str, help='Config property to retrieve.')
 
+    #  subparser for sa-conf-delete
+    parser_account = subparsers.add_parser('sa-conf-delete')
+    parser_account.add_argument('--service-account-name', default=None, help='Service Account for which configuration properties need to be deleted.')
+
     args = parser.parse_args()
 
     defaults = get_defaults_from_kubeconfig(args.kubeconfig or DEFAULT_KUBECONFIG, args.context)
@@ -154,4 +158,6 @@ if __name__ == "__main__":
         utils.setup_kubernetes_secret(args.properties_file, args.service_account_name)
     elif args.action == 'sa-conf-get':
         utils.retrieve_kubernetes_secret(args.service_account_name, args.conf)
+    elif args.action == 'sa-conf-delete':
+        utils.delete_kubernetes_secret(args.service_account_name)
 
--- a/helpers/setup-spark-k8s.py
+++ b/helpers/setup-spark-k8s.py
@@ -1,151 +1,12 @@
 #!/usr/bin/env python3
 
-import sys
-import os
-import yaml
 import argparse
-import pwd
-from typing import Dict
 import utils
 import logging
-
-KUBECTL_CMD= '{}/kubectl'.format(os.environ['SNAP'])
-USER_HOME_DIR_ENT_IDX = 5
-
-def print_help_for_missing_or_inaccessible_kubeconfig_file(kubeconfig: str):
-    print('\nERROR: Missing kubeconfig file {}. Or default kubeconfig file {}/.kube/config not found.'.format(kubeconfig, pwd.getpwuid(os.getuid())[5]))
-    print('\n')
-    print('Looks like either kubernetes is not set up properly or default kubeconfig file is not accessible!')
-    print('Please take the following remedial actions.')
-    print('	1. Please set up kubernetes and make sure kubeconfig file is available, accessible and correct.')
-    print('	2. sudo snap connect spark-client:dot-kube-config')
-
-def print_help_for_bad_kubeconfig_file(kubeconfig: str):
-    print('\nERROR: Invalid or incomplete kubeconfig file {}. One or more of the following entries might be missing or invalid.\n'.format(kubeconfig))
-    print('	- current-context')
-    print('	- context.name')
-    print('	- context.namespace')
-    print('	- context.cluster')
-    print('	- cluster.name')
-    print('	- cluster.server')
-    print(' - cluster.certificate-authority-data')
-    print('Please take the following remedial actions.')
-    print('	1. Please set up kubernetes and make sure kubeconfig file is available, accessible and correct.')
-    print('	2. sudo snap connect spark-client:dot-kube-config')
-
-def select_context_id(kube_cfg: Dict) -> int:
-    NO_CONTEXT = -1
-    SINGLE_CONTEXT = 0
-    context_names = [n['name'] for n in kube_cfg['contexts']]
-
-    selected_context_id = NO_CONTEXT
-    if len(context_names) == 1:
-        return SINGLE_CONTEXT
-
-    context_list_ux = '\n'.join(["{}. {}".format(context_names.index(a), a) for a in context_names])
-    while selected_context_id == NO_CONTEXT:
-        print (context_list_ux)
-        print('\nPlease select a kubernetes context by index:')
-        selected_context_id = input()
-
-        try:
-           int(selected_context_id)
-        except ValueError:
-            print('Invalid context index selection, please try again....')
-            selected_context_id = NO_CONTEXT
-            continue
-
-        if int(selected_context_id) not in range(len(context_names)):
-            print('Invalid context index selection, please try again....')
-            selected_context_id = NO_CONTEXT
-            continue
-
-    return int(selected_context_id)
-
-def get_defaults_from_kubeconfig(kubeconfig: str, context: str = None) -> Dict:
-    try:
-        with open(kubeconfig) as f:
-            kube_cfg = yaml.safe_load(f)
-            context_names = [n['name'] for n in kube_cfg['contexts']]
-            certs = [n['cluster']['certificate-authority-data'] for n in kube_cfg['clusters']]
-            current_context = context or kube_cfg['current-context']
-    except IOError as ioe:
-        print_help_for_missing_or_inaccessible_kubeconfig_file(kubeconfig)
-        sys.exit(-1)
-    except KeyError as ke:
-        print_help_for_bad_kubeconfig_file(kubeconfig)
-        sys.exit(-2)
-
-    try:
-        context_id = context_names.index(current_context)
-    except ValueError:
-        print(f'WARNING: Current context in provided kubeconfig file {kubeconfig} is invalid!')
-        print(f'\nProceeding with explicit context selection....')
-        context_id = select_context_id(kube_cfg)
-
-    defaults = {}
-    defaults['context'] = context_names[context_id]
-    defaults['namespace'] = 'default'
-    defaults['cert'] = certs[context_id]
-    defaults['config'] = kubeconfig
-    defaults['user'] = 'spark'
-
-    return defaults
-
-def set_up_user(username: str, name_space: str, defaults: Dict, mark_primary: bool) -> None:
-    namespace = name_space or defaults['namespace']
-    kubeconfig = defaults['config']
-    context_name = defaults['context']
-
-    rolebindingname = username + '-role'
-    roleaccess='view'
-    label = utils.get_management_label()
-
-    os.system(f"{KUBECTL_CMD} create serviceaccount --kubeconfig={kubeconfig} --context={context_name} {username} --namespace={namespace}")
-    os.system(f"{KUBECTL_CMD} create rolebinding --kubeconfig={kubeconfig} --context={context_name} {rolebindingname} --role={roleaccess}  --serviceaccount={namespace}:{username} --namespace={namespace}")
-
-    primary_label_to_remove = utils.get_primary_label(label=False)
-    primary_label_full = utils.get_primary_label()
-
-    primary = utils.retrieve_primary_service_account_details(namespace, context_name)
-    is_primary_defined = len(primary.keys()) > 0
-
-    logging.debug(f"is_primary_defined={is_primary_defined}")
-    logging.debug(f"mark_primary={mark_primary}")
-
-    if is_primary_defined and mark_primary:
-        sa_to_unlabel = primary['spark.kubernetes.authenticate.driver.serviceAccountName']
-        namespace_of_sa_to_unlabel = primary['spark.kubernetes.namespace']
-        rolebindingname_to_unlabel = sa_to_unlabel + '-role'
-
-        os.system(f"{KUBECTL_CMD} label serviceaccount --kubeconfig={kubeconfig} --context={context_name} --namespace={namespace_of_sa_to_unlabel} {sa_to_unlabel} {primary_label_to_remove}-")
-        os.system(f"{KUBECTL_CMD} label rolebinding --kubeconfig={kubeconfig} --context={context_name} --namespace={namespace_of_sa_to_unlabel} {rolebindingname_to_unlabel} {primary_label_to_remove}-")
-
-        os.system(f"{KUBECTL_CMD} label serviceaccount --kubeconfig={kubeconfig} --context={context_name} {username} {label} {primary_label_full} --namespace={namespace}")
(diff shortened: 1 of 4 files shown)


helpers/utils.py
```diff
@@ -107,4 +118,47 @@ def UmaskNamedTemporaryFile(*args, **kargs):
...
+        subprocess.check_output(cmd, shell=True)
+    except subprocess.CalledProcessError as e:
+        logging.error(
+            f'Cannot create secret. Probably bad or missing kubeconfig file {kubeconfig}')
+        logging.warning(
+            f'Please set the KUBECONFIG environment variable to point to the kubeconfig. Or fix the default kubeconfig {kubeconfig}')
+        logging.error(e.output)
+        sys.exit(EXIT_CODE_BAD_KUBECONFIG)
+
+def retrieve_kubernetes_secret(service_account_name: str, keys: List[str]) -> None:
```
averma-canonical, line 140: namespace

helpers/utils.py
```diff
@@ -107,4 +118,47 @@ def UmaskNamedTemporaryFile(*args, **kargs):
...
+            out = subprocess.check_output(cmd, shell=True)
+            secret = yaml.safe_load(out.decode("utf-8"))
+            keys = secret['data'].keys()
+        except subprocess.CalledProcessError as e:
+            logging.error(e.output)
+            sys.exit(EXIT_CODE_GET_SECRET_FAILED)
+
+    for k in keys:
+        k1 = k.replace('.', '\\.')
+        cmd = f"{kubectl_cmd} --kubeconfig {kubeconfig} get secret {secret_name} -o jsonpath='{{.data.{k1}}}' | base64 --decode"
```
averma-canonical, line 157: Do we need to get the kubernetes context from user as well or we don't really care?

helpers/utils.py
```diff
@@ -107,4 +118,47 @@ def UmaskNamedTemporaryFile(*args, **kargs):
...
+        try:
+            out = subprocess.check_output(cmd, shell=True)
+            v = out.decode("utf-8")
+            print(f"{k}={v}")
+        except subprocess.CalledProcessError as e:
+            logging.error(e.output)
+            sys.exit(EXIT_CODE_GET_SECRET_FAILED)
```
averma-canonical, line 164: delete_kubernetes_secret() TBD.

welpaolo (commented):
Hi Abhishek, the code looks ok. 
I have two main comments: 
- Can you please add docstring for each function to describe its purpose?
- I would add some inline comments to explain the more complex commands or iteration (e.g., for function `retrieve_kubernetes_secret`)

Do you plan to switch to the lightkube library now or in some future PRs ?

helpers/setup-spark-k8s.py
```diff
@@ -139,3 +154,10 @@ def setup_spark_conf_defaults(username: str, namespace: str) -> None:
         namespace = args.namespace or defaults['namespace']
         set_up_user(username, namespace, defaults)
         setup_spark_conf_defaults(username, namespace)
+    elif args.action == 'sa-conf-create':
+        utils.setup_kubernetes_secret(args.properties_file, args.service_account_name)
```
deusebio, line 158: What about creating the properties when we setup the service account ? We could provide some flags (as properties file or a list of --conf flags) to the "setup-spark-k8s service account" command

helpers/utils.py
```diff
@@ -107,4 +118,47 @@ def UmaskNamedTemporaryFile(*args, **kargs):
     umask = os.umask(0o666)
     os.umask(umask)
     os.chmod(fdesc.name, 0o666 & ~umask)
-    return fdesc
\ No newline at end of file
+    return fdesc
+
+def setup_kubernetes_secret(properties_file: str, service_account_name: str) -> None:
```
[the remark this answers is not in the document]
deusebio: I was about to say... :D

helpers/utils.py
```diff
@@ -107,4 +119,58 @@ def UmaskNamedTemporaryFile(*args, **kargs):
...
+            f'Please set the KUBECONFIG environment variable to point to the kubeconfig. Or fix the default kubeconfig {kubeconfig}')
+        logging.error(e.output)
+        sys.exit(EXIT_CODE_BAD_KUBECONFIG)
+
+def retrieve_kubernetes_secret(service_account_name: str, keys: List[str]) -> None:
+    secret_name = f"spark-client-sa-conf-{service_account_name or 'spark'}"
+    kubeconfig = get_kube_config()
+    kubectl_cmd = get_kubectl_cmd()
+
+    if not keys or len(keys) == 0:
```
deusebio, line 146: I would break the business logic into parsing the properties AND printing the values. 

Note that printing the values could be common to any property information (provided usually as dict), whereas reasing properties from k8s is generally what we would like to do when generating the dynamic configuration

deusebio:
We have not yet done the dynamic creation of the properties, have we? It seems that we are only writing data to k8s and not ever reading it for injecting that into the spark-submit command

Another point: we should also implement the logic to automatically retrieve the serviceaccount / namespace using the label, as Mykola suggested. I don't think this business logic is yet implemented, isn't it?

helpers/spark-submit.py
```diff
@@ -24,16 +24,19 @@
...
     STATIC_DEFAULTS_CONF_FILE = utils.get_static_defaults_conf_file()
-    DYNAMIC_DEFAULTS_CONF_FILE = utils.get_dynamic_defaults_conf_file()
     ENV_DEFAULTS_CONF_FILE = utils.get_env_defaults_conf_file()
+    primary = utils.retrieve_primary_service_account_details()
+    primary_username = primary['spark.kubernetes.authenticate.driver.serviceAccountName']
+    primary_namespace = primary['spark.kubernetes.namespace']
 
     snap_static_defaults = utils.read_property_file(STATIC_DEFAULTS_CONF_FILE)
-    setup_dynamic_defaults = utils.read_property_file(DYNAMIC_DEFAULTS_CONF_FILE) if os.path.isfile(DYNAMIC_DEFAULTS_CONF_FILE) else dict()
+    setup_dynamic_defaults = utils.retrieve_primary_service_account_details()
```
deusebio, line 33: why are we calling retrieve_primary_service_account_details twice? line 33 and line 28?

To be honest for better readability of the code, I would really just compress this business logic into a single function 

`dynamic_defautls = utils.get_dynamic_defaults(service_accont)`

with

`def get_dynamic_defaults(service_account: Optional[str]) -> Dict`

I put this as optional since, following the ticket description, one could also provide a service account via CLI --service-account=namespace:service-account. If no service account is provided, we actually do the inference that you are doing here, otherwise we use the one provided by the user

helpers/utils.py
```diff
@@ -9,10 +9,18 @@
...
 USER_HOME_DIR_ENT_IDX = 5
 EXIT_CODE_BAD_KUBECONFIG = -100
 EXIT_CODE_BAD_CONF_ARG = -200
+EXIT_CODE_GET_SECRET_FAILED = -300
+EXIT_CODE_PUT_SECRET_ENV_FILE_FAILED = -400
+EXIT_CODE_PUT_SECRET_LITERAL_FAILED = -500
+EXIT_CODE_DEL_SECRET_FAILED = -600
+EXIT_CODE_GET_K8S_PROPS_FAILED = -700
+EXIT_CODE_GET_PRIMARY_RESOURCES_FAILED = -800
+EXIT_CODE_SET_PRIMARY_RESOURCES_FAILED = -900
```
deusebio, line 23: We should probably build an enum object with all of these, for better readability and grouping. But now that I'm thinking, we should also probably tackle these with exceptions and exception handlers. 

Anyhow, it is probably not worth addressing this in this PR (and it goes beyond this scope), but we need to do some thinking and design on how we want to handle exceptions at scale. Just mentioning here, so that we start to think about this improvement

helpers/setup-spark-k8s.py
```diff
@@ -91,44 +92,94 @@ def get_defaults_from_kubeconfig(kubeconfig: str, context: str = None) -> Dict:
...
-    parser_account.add_argument('--namespace', default=None, help='Namespace for the service account to be created in kubernetes. Default is default namespace')
+    parser_account.add_argument('--primary', action='store_true', help='Boolean to mark the service account as primary.')
+    parser_account.add_argument('--conf', action='append', type=str, help='Config properties to be added to the service account.')
+
+    #  subparser for service-account-cleanup
+    parser_account = subparsers.add_parser('service-account-cleanup')
+
+    #  subparser for sa-conf-put
+    parser_account = subparsers.add_parser('sa-conf-put')
+    parser_account.add_argument('--properties-file', default = None, help='File with all configuration properties assignments.')
```
deusebio, line 169: why not  giving the "--conf" to this api?

helpers/setup-spark-k8s.py
```diff
@@ -91,44 +92,94 @@ def get_defaults_from_kubeconfig(kubeconfig: str, context: str = None) -> Dict:
...
+
     subparsers = parser.add_subparsers(dest='action')
     subparsers.required = True
 
     #  subparser for service-account
     parser_account = subparsers.add_parser('service-account')
-    parser_account.add_argument('--username', default='spark', help='Service account username to be created in kubernetes. Default is spark')
-    parser_account.add_argument('--namespace', default=None, help='Namespace for the service account to be created in kubernetes. Default is default namespace')
+    parser_account.add_argument('--primary', action='store_true', help='Boolean to mark the service account as primary.')
+    parser_account.add_argument('--conf', action='append', type=str, help='Config properties to be added to the service account.')
```
deusebio, line 162: why not giving the "--properties-file" to this api?

helpers/setup-spark-k8s.py
```diff
@@ -137,5 +188,21 @@ def setup_spark_conf_defaults(username: str, namespace: str) -> None:
...
-        set_up_user(username, namespace, defaults)
-        setup_spark_conf_defaults(username, namespace)
+        set_up_user(username, namespace, defaults, args.primary)
+        utils.setup_kubernetes_secret_literal(username, namespace, args.context, args.conf)
+    elif args.action == 'service-account-cleanup':
+        cleanup_user(args.username, args.namespace, args.context)
+    elif args.action == 'sa-conf-put':
+        utils.setup_kubernetes_secret_env_file(args.username, args.namespace or defaults['namespace'], args.context, args.properties_file)
+    elif args.action == 'sa-conf-get':
+        conf = utils.retrieve_kubernetes_secret(args.username, args.namespace, args.context, args.conf)
```
deusebio, line 198: why having two different UX for setting up account and sa-conf-put?

I would just allow both to take either --conf or --properties-file (with --conf overriding --properties-file if the same key is provided). It would be good to have the same function in both places

helpers/utils.py
```diff
@@ -107,4 +128,117 @@ def UmaskNamedTemporaryFile(*args, **kargs):
...
+
+def build_kubectl_cmd(namespace: str, k8s_context: str) -> str:
+    kubeconfig = get_kube_config()
+    kubectl_cmd = get_kubectl_cmd()
+    cmd = f"{kubectl_cmd} --kubeconfig {kubeconfig} --namespace={namespace}"
+    if k8s_context:
+        cmd += f" --context {k8s_context}"
+    return cmd
+
+def build_secret_name(username: str, namespace: str) -> str:
```
deusebio, line 141: you don't need namespace, do you?

username should be Optional (inferring this from your code "username or spark")

helpers/utils.py
```diff
@@ -107,4 +128,117 @@ def UmaskNamedTemporaryFile(*args, **kargs):
...
+    defaults = generate_spark_default_conf()
+    for k in defaults.keys():
+        cmd += f" --from-literal={k}={defaults[k]}"
+
+    cmd += f" --from-literal=spark.kubernetes.authenticate.driver.serviceAccountName={username}"
+    cmd += f" --from-literal=spark.kubernetes.namespace={namespace}"
+
+    execute_kubectl_cmd(cmd, EXIT_CODE_PUT_SECRET_LITERAL_FAILED)
+
+def setup_kubernetes_secret_env_file(username: str, namespace: str, k8s_context: str, properties_file: str) -> None:
```
deusebio, line 178: In order to merge the behaviour, I would really build this on top of features we have already implemented:

1. Parse --conf into a dict
2. Read the properties-file into a dict
3. merge the two dicts
4. write this properties-file to a temporary file
5. submit secret creation via kubectl
6. remove temporary file

helpers/utils.py
```diff
@@ -107,4 +128,117 @@ def UmaskNamedTemporaryFile(*args, **kargs):
...
+    if conf:
+        for c in conf:
+            cmd += f" --from-literal={c}"
+
+    defaults = generate_spark_default_conf()
+    for k in defaults.keys():
+        cmd += f" --from-literal={k}={defaults[k]}"
+
+    cmd += f" --from-literal=spark.kubernetes.authenticate.driver.serviceAccountName={username}"
+    cmd += f" --from-literal=spark.kubernetes.namespace={namespace}"
```
deusebio, line 174: No need to create these. These will be automatically generated and merged in spark-submit.

helpers/setup-spark-k8s.py
```diff
@@ -91,44 +92,94 @@ def get_defaults_from_kubeconfig(kubeconfig: str, context: str = None) -> Dict:
...
+
+    logging.debug(f"is_primary_defined={is_primary_defined}")
+    logging.debug(f"mark_primary={mark_primary}")
+
+    if is_primary_defined and mark_primary:
+        sa_to_unlabel = primary['spark.kubernetes.authenticate.driver.serviceAccountName']
+        namespace_of_sa_to_unlabel = primary['spark.kubernetes.namespace']
+        rolebindingname_to_unlabel = sa_to_unlabel + '-role'
+
+        os.system(f"{KUBECTL_CMD} label serviceaccount --kubeconfig={kubeconfig} --context={context_name} --namespace={namespace_of_sa_to_unlabel} {sa_to_unlabel} {primary_label_to_remove}-")
```
deusebio, line 121: Sometimes we use KUBECTL_CMD and sometimes we have a helpers function get_kubectl_cmd, or build_kubectl_cmd. 

When do we use one or another? Do we have use-cases for having multiple ways to do similar things? Can we think of refactors that would allow us to centralize behaviour and implementation, in order to promove code reusability and consistent API/UX?

deusebio (commented):
I believe there are still a few items we need to address. 

After this pass, I'd like to start play around with it

helpers/utils.py
```diff
@@ -75,36 +113,401 @@ def parse_conf_overrides(conf_args: List) -> Dict:
...
-       sys.exit(EXIT_CODE_BAD_KUBECONFIG)
-    return k8s_master_uri
+    primary_sa = retrieve_primary_service_account_details(namespace, kubeconfig, context)
+    primary_namespace = primary_sa.get('spark.kubernetes.namespace') or 'default'
+    kubectl_cmd = build_kubectl_cmd(kubeconfig, primary_namespace, context)
+    master_cmd = f"{kubectl_cmd} config view --minify -o jsonpath=\"{{.clusters[0]['cluster.server']}}\""
+    default_master = execute_kubectl_cmd(master_cmd, EXIT_CODE_BAD_KUBECONFIG)
+    return f'k8s://{default_master}'
 
 def UmaskNamedTemporaryFile(*args, **kargs):
```
deusebio, line 157: I know it does not pertain this PR...but I suppose we should go for the lower case, snake convention here, being a python function

deusebio (approved):
Looks good! We need to do a bit of restructuring and refactoring of the helpers. But API is really starting to shape well and neat! 

Nice!

helpers/utils.py
averma-canonical: noted.

PR #7, opened 2022-11-09
SNAP Structuring Proposal

deusebio:


helpers/utils.py
```diff
@@ -175,70 +215,126 @@ def setup_kubernetes_secret_literal(username: str, namespace: str, k8s_context:
...
+
+
+def mkdir(path: PathLike) -> None:
+    """
+    Create a dir, using a formulation consistent between 2.x and 3.x python versions.
+    :param path: path to create
+    :raises OSError: whenever OSError is raised by makedirs and it's not because the directory exists
+    """
+    try:
+        os.makedirs(path)
```
averma-canonical, line 324: Thanks for adding these functions, were these added automatically?  There might be more such utilities with different behaviors in 2.x and 3.x. Let's discuss.

deusebio: Copied and paste from a repo I had. To be honest, given it is 2022 (almost 2023) and we don't have the need to support backward compatibility to 2.x, I would really just drop support for older version of python. I'd stick to ">3.9" (I have updated updated CI to reflect this), but I don't have super strong opinion on the actual version we want to start comply with. Just 2.x feels very very very very old... (any strong opinion by @delgod or @taurus-forever would be much appreciated)

I believe it would be eventually good to have core utilities function, if we see the use case for it, to promote patterns, best-practices, and so, even in a form of a python package where we centralize these code. But I suppose I'd wait a bit to understand if we have such use-cases. So far, I just rely on copy and paste ;)

averma-canonical (approved):
Looks great. Thanks for adding "--log-level" option to spark-submit and setup-spark-k8s.

.github/workflows/snap.yaml
```diff
@@ -91,58 +107,14 @@ jobs:
...
-      - name: Setup client
-        run: |
-          # create the service account
-          spark-client.setup-spark-k8s service-account
+      - name: Checkout repo
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ env.BRANCH }}
 
       - name: Run example job
```
welpaolo, line 115: Maybe `Run integration tests` will be a better label

welpaolo (approved):
Look good to me. Just left a minor comment

deusebio:
Addressed @welpaolo comment. But as I re-pushed a new version, re-approvals are required! 

@averma-canonical @welpaolo : Please do provide approvals again

averma-canonical (approved):
Looks good.

welpaolo (dismissed):
Looks good to me

tests/integration/ie_tests.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
delgod, line 1: I don't think that `sudo` is needed in this file.
We should not require root access for communication with kubernetes.

deusebio: Done!

delgod (approved):

deusebio:
@delgod : Making it work without `sudo` was not as straightforward as I thought, due to the fact that `kubectl` can be used only for microk8s group users. However the user running the script does not have the permission in its current session and it would be granted the microk8s permission only after re-logging. 

In the end I had to rely on running scripts using `sg microk8s my-script.sh`

@delgod @welpaolo @averma-canonical   Just double check this makes sense and the modifications are fine. Please do re-approve in this case

tox.ini
```diff
@@ -0,0 +1,54 @@
...
+    poetry install --with dev --sync
+    poetry export -f requirements.txt -o requirements.txt
+    poetry run pytest tests/unittest
+
+[testenv:integration]
+description = Run integration tests
+commands =
+    poetry install --with dev --sync
+    poetry export -f requirements.txt -o requirements.txt
+    sg microk8s {toxinidir}/tests/integration/ie-tests.sh
```
deusebio, line 54: see here

welpaolo (approved):

averma-canonical (approved):

PR #8, opened 2022-11-10
DPE-839: Introducing python wrapper for spark-shell within the snap.

averma-canonical:
We want to create a wrapper for the “spark-shell” command, **building on the parsing of configurations and merging logic already implemented for spark-submit**

Main features:
1. deploy-mode: set to “client” in any case (as spark-shell should really not be run with deploy-mode set to “cluster”)
2. setting the .scala_history file location dynamically
- $SNAP_USER_DATA/.scala_history if SNAP_USER_DATA defined OR
- $HOME/.scala_history

### DIFF
--- a/.github/workflows/snap.yaml
+++ b/.github/workflows/snap.yaml
@@ -148,7 +148,7 @@ jobs:
           echo "val count = spark.sparkContext.parallelize(1 until n, slices).map { i => val x = random * 2 - 1; val y = random * 2 - 1;  if (x*x + y*y <= 1) 1 else 0;}.reduce(_ + _)" >> test-spark-shell.scala
           echo "println(s\"Pi is roughly \${4.0 * count / (n - 1)}\")" >> test-spark-shell.scala
           echo "System.exit(0)" >> test-spark-shell.scala
-          spark-client.spark-shell  --conf spark.kubernetes.container.image=averma32/sparkrock:latest < test-spark-shell.scala > spark-shell.out
+          spark-client.spark-shell < test-spark-shell.scala > spark-shell.out
           pi=$(grep -v "scala>" spark-shell.out  | grep "Pi is roughly" | rev | cut -d' ' -f1 | rev | cut -c 1-4)
           echo -e "Spark-shell Pi Job Output: \n ${pi}"
           if [ "${pi}" != "3.14" ]; then
--- a/.github/workflows/snap.yaml
+++ b/.github/workflows/snap.yaml
@@ -149,7 +149,7 @@ jobs:
           echo "val count = spark.sparkContext.parallelize(1 until n, slices).map { i => val x = random * 2 - 1; val y = random * 2 - 1;  if (x*x + y*y <= 1) 1 else 0;}.reduce(_ + _)" >> test-spark-shell.scala
           echo "println(s\"Pi is roughly \${4.0 * count / (n - 1)}\")" >> test-spark-shell.scala
           echo "System.exit(0)" >> test-spark-shell.scala
-          spark-client.spark-shell < test-spark-shell.scala > spark-shell.out
+          spark-client.spark-shell --conf spark.kubernetes.container.image=docker.io/averma32/sparkpy6:latest < test-spark-shell.scala > spark-shell.out
           pi=$(grep -v "scala>" spark-shell.out  | grep "Pi is roughly" | rev | cut -d' ' -f1 | rev | cut -c 1-4)
           echo -e "Spark-shell Pi Job Output: \n ${pi}"
           if [ "${pi}" != "3.14" ]; then
--- a/.github/workflows/snap.yaml
+++ b/.github/workflows/snap.yaml
@@ -142,14 +142,14 @@ jobs:
 
       - name: Test spark-shell
         run: |
-          echo "import scala.math.random" > /home/runnner/test-spark-shell.scala
-          echo "val slices = 1000" >> /home/runnner/test-spark-shell.scala
-          echo "val n = math.min(100000L * slices, Int.MaxValue).toInt" >> /home/runnner/test-spark-shell.scala
-          echo "val count = spark.sparkContext.parallelize(1 until n, slices).map { i => val x = random * 2 - 1; val y = random * 2 - 1;  if (x*x + y*y <= 1) 1 else 0;}.reduce(_ + _)" >> /home/runnner/test-spark-shell.scala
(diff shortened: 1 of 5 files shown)


helpers/spark-shell.py
```diff
@@ -0,0 +1,50 @@
...
+import utils
+
+USER_HOME_DIR_ENT_IDX = 5
+
+if __name__ == "__main__":
+    logging.basicConfig(format='%(asctime)s %(levelname)s %(message)s', level=logging.INFO)
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--master", default=None, type=str, help='Kubernetes control plane uri.')
+    parser.add_argument("--deploy-mode", default="client", type=str, help='Deployment mode for spark shell. Default is \'client\'.')
```
deusebio, line 16: Uhm, not sure we should allow the user to change this. Beside whatever he inputs, it would not be taken from the application. So, I'd just skip this argument

helpers/spark-shell.py
```diff
@@ -0,0 +1,50 @@
...
+
+    snap_static_defaults = utils.read_property_file(STATIC_DEFAULTS_CONF_FILE)
+    setup_dynamic_defaults = utils.read_property_file(DYNAMIC_DEFAULTS_CONF_FILE) if os.path.isfile(DYNAMIC_DEFAULTS_CONF_FILE) else dict()
+    env_defaults = utils.read_property_file(ENV_DEFAULTS_CONF_FILE) if ENV_DEFAULTS_CONF_FILE and os.path.isfile(ENV_DEFAULTS_CONF_FILE) else dict()
+    props_file_arg_defaults = utils.read_property_file(args.properties_file) if args.properties_file else dict()
+
+    with utils.UmaskNamedTemporaryFile(mode = 'w', prefix='spark-conf-', suffix='.conf') as t:
+        defaults = utils.merge_configurations([snap_static_defaults, setup_dynamic_defaults, env_defaults, props_file_arg_defaults])
+        SCALA_HISTORY_FILE = utils.get_scala_shell_history_file()
+        defaults['spark.driver.extraJavaOptions'] = f'-Dscala.shell.histfile={SCALA_HISTORY_FILE}'
```
deusebio, line 36: what does it happen if this key is already present? Can we put some logic to merge the two configurations for the extraJavaOptions rather than overwriting them?

I suppose we could see the use case where the user specifies some parameters that then gets overridden by this line. 

BTW...it would be nice to carry that merging logic also to the other parts of the merge_configurations. But let's discuss this a bit more before addressing this. Here I'm just mentioning to start to think about this

helpers/utils.py
```diff
@@ -57,6 +57,12 @@ def get_env_defaults_conf_file() -> str:
     SPARK_ENV_DEFAULTS_FILE = os.environ.get('SNAP_SPARK_ENV_CONF')
     return SPARK_ENV_DEFAULTS_FILE
 
+def get_scala_shell_history_file() -> str:
+    """Returns location of .scala_history file for spark-shell"""
+    USER_HOME_DIR = pwd.getpwuid(os.getuid())[USER_HOME_DIR_ENT_IDX]
+    SCALA_HIST_FILE_DIR = os.environ.get('SNAP_USER_DATA', USER_HOME_DIR)
+    return f'{SCALA_HIST_FILE_DIR}/.scala_history'
```
deusebio, line 64: Should we allow the user to specify the name of this file directly via an env variable?

deusebio (commented):
Looks good!

I only have some minor comments attached worth being address before merging

helpers/spark-shell.py
averma-canonical: Actually, if we do not explicitly parse this out, it would go to un-parsed extra_args where it would be more difficult to change it from cluster to client for example. hence, I thought of parsing it away and then overriding it explicitly to client every time.

If you notice below, I never actually consult or use args.deploy_mode at all in the final spark-shell command.

helpers/spark-shell.py
averma-canonical: Yeah, this does not look right. 

Instead, I guess I can do this immediately after getting setup related dynamic defaults.

dynamic_defaults['spark.driver.extraJavaOptions'] = f'-Dscala.shell.histfile={SCALA_HISTORY_FILE}'

This way it can be overridden with env file and properties file that are parsed next in order.

helpers/utils.py
averma-canonical: Not sure if we need an extra env variable for just this property. 

env file and --properties-file can always include this property. Please check my suggestion above where we add our default value of scala_history file location to setup related dynamic defaults, which are parsed before env file and --properties-file.

.github/workflows/snap.yaml
```diff
@@ -140,6 +141,21 @@ jobs:
...
+          echo "val n = math.min(100000L * slices, Int.MaxValue).toInt" >> test-spark-shell.scala
+          echo "val count = spark.sparkContext.parallelize(1 until n, slices).map { i => val x = random * 2 - 1; val y = random * 2 - 1;  if (x*x + y*y <= 1) 1 else 0;}.reduce(_ + _)" >> test-spark-shell.scala
+          echo "println(s\"Pi is roughly \${4.0 * count / (n - 1)}\")" >> test-spark-shell.scala
+          echo "System.exit(0)" >> test-spark-shell.scala
+          spark-client.spark-shell < test-spark-shell.scala > spark-shell.out
+          pi=$(grep -v "scala>" spark-shell.out  | grep "Pi is roughly" | rev | cut -d' ' -f1 | rev | cut -c 1-4)
+          echo -e "Spark-shell Pi Job Output: \n ${pi}"
+          if [ "${pi}" != "3.14" ]; then
+              exit 1
+          fi
```
deusebio, line 157: We will have to move this to the ie_tests in the new skeleton

helpers/spark-defaults.conf
```diff
@@ -1 +1 @@
-spark.kubernetes.container.image=docker.io/averma32/sparkpy6:latest
\ No newline at end of file
+spark.kubernetes.container.image=docker.io/averma32/sparkrock:latest
```
deusebio, line 1: very well! :+1:

deusebio (dismissed):
It looks very good! Just wait the other PR to be merged and please do refactor this by moving the test into the proper place such that it will be run automatically, either with makefile or during CI/CD (which uses the same entrypoint as Makefile)

averma-canonical:
Cancelling this PR and created a new one after rebasing with Enrico's new test framework.

PR #10, opened 2022-11-18
DPE-902: Updating docs with info on service account management and configuration resolution during spark-submit.

averma-canonical:
Added reference documentation for 
* service account and config properties management utils in setup command.
* configuration resolution behavior at runtime for job submissions


docs/setup.md
```diff
@@ -50,16 +64,93 @@ To get help regarding the usage of service account setup command within the snap
...
+
+#### Deleting configuration for a Service Account
+
+For deleting all configuration associated with a service account,
+
+```bash
+spark-client.setup-spark-k8s --username demouser --namespace demonamespace sa-conf-delete
+```
+
+#### Inspecting Primary Service Account
```
averma-canonical, line 138: Feels like following functionalities would be useful 
-- mark an existing service account as primary
-- list all service accounts managed by Spark client
-- register an existing service account for use with Spark client (will be useful when people using kubeflow like products just want to install the client and go with the existing service account which they might have created in a very custom way which cannot be repeated with Spark client.)

docs/config.md
```diff
@@ -0,0 +1,21 @@
+### Configuration Setup and Runtime behavior for Apache Spark client
+Apache Spark comes with a lots of configuration properties which can be set. The list is pretty extensible as well.
```
deusebio, line 2: Should we provide the link?

docs/config.md
```diff
@@ -0,0 +1,21 @@
...
+
+Passing each and every configuration in command line is not tenable, so Apache Spark supports the use of properties 
+files where the user can place most of the configuration for reuse across submissions. 
+
+In addition, the user can still add or override more configuration values in this file in the command line.
+
+For this Spark client, Canonical has come up with a rich set of options for the user to specify configuration properties and also override
+them in a spirit similar to Apache Spark.
```
deusebio, line 10:
```suggestion
For this Spark client, Canonical has come up with a rich set of options for the user to specify configuration properties and also override
them in a spirit similar to Apache Spark.
```

Spark client provides a rich set of options for the user to specify configuration properties and also override
them in a spirit similar to Apache Spark.

docs/config.md
```diff
@@ -0,0 +1,21 @@
...
+
+Following is the hierarchy of configurations merged during spark-submit
+* Canonical provided immutable defaults
+* Set up time generated immutable defaults kept as a secret collection in Kubernetes. Valid across sessions and machines. Please refer to the [setup](/docs/setup.md) section, specifically the part about ```service-account```. 
+* Properties in file specified via environment variable (```$SNAP_SPARK_ENV_CONF```) valid across spark-submit commands in a shell session.
+* Properties in file specified as a parameter (```--properties-file```) in spark-submit command
+* Properties specified as parameters (list of ```--conf```) in spark-submit command.
```
deusebio, line 17: I would give "nicknames" to these properties, to also have a way to easily reference them in the text:

* SNAP Configuration
* Service Account Configuration
* Environment Configuration
* CLI Configuration (provided either via --properties-file or --conf)

docs/setup.md
deusebio: Yes, I agree. They are all very relevant points. I would put them in the backlog and see whether we will have margin to address (some or all of) them in the next sprints!

docs/setup.md
```diff
@@ -50,16 +64,93 @@ To get help regarding the usage of service account setup command within the snap
...
+
+For example, assuming the properties file provided has configuration details to access data in S3, one could submit a job like
+```bash
+spark-client.spark-submit  --username demouser --namespace demonamespace --deploy-mode cluster --conf spark.app.name=demo-spark-s3-app $S3_PATH_FOR_CODE_FILE
+```
+This would launch the spark job with configuration coming from the service account for user ```demonamespace:demouser``` but the app name would be ```demo-spark-s3-app```. 
+
+**_NOTE:_** The comamnd described above does not create a kubernetes namespace but needs it to be there. It does however create the requested username in the specified and existing namespace. 
+
+#### Creating configuration for a Service Account after account creation
```
deusebio, line 103: These look more like a "How to Guide". I would probably just make a section called "How To" , where we put these commands

deusebio (changes requested):
In the Jira ticket we wrote to write a tutorial, but when reading through the docs, I believe we ended up more with a "How-to" guide, which is perferctly fine too. I would just make it clearer that these is a How-To. 

Just to be on the same page, In a tutorial we should rather tell a story, probably creating more accounts, deleting some of them, launch some jobs. In a tutorial, we should probably also write the output that one expects. Anyhow, It can be something for another ticket on documentation. 

So no need to do extra work apart from rephrasing things as How-to

docs/setup.md
averma-canonical: Done

README.md
```diff
@@ -12,13 +12,17 @@ This repository hosts the source for ***spark-client*** snap from Canonical. The
 The spark-client snap from Canonical simplifies the setup to run Spark jobs against your Kubernetes cluster. Please follow the
```
welpaolo, line 12: I don't know if we can/want to mention Canonical here. (for legal reasons)

welpaolo (approved):
I just left a minor comment. The rest of the documentation look good to me

docs/howto.md
```diff
@@ -0,0 +1,54 @@
+#### How to create configuration for a Service Account after account creation
```
deusebio, line 1: Just create a section 

### How-To Guide

This section provides general commands on how to use the CLI interface for creating, managing and configuration spark applications. 

##### Create Service Account 

```bash
....
````

If the account "demouser" already exists, this will drop

File: docs/howto.md [hunk 8]
```diff
@@ -0,0 +1,54 @@
...
+
+The configuration functionality mentioned in the previous section can be done after account creation as well. Useful for redefining the configuration associated with a existing service account.
```
deusebio, line 3: We should not refer if possible other material. This should be self-contained

docs/howto.md
```diff
@@ -0,0 +1,54 @@
...
+
+For deleting all configuration associated with a service account,
+
+```bash
+spark-client.setup-spark-k8s --username demouser --namespace demonamespace sa-conf-delete
+```
+
+#### How to check Primary Service Account
+
+While discussing service account creation, it was mentioned that a service account can be marked as the ```primary``` account to be used implicitly if no service account is specified during job submission.
```
deusebio, line 38: This is more explanation

deusebio (changes requested):
Sorry for being pedantic here, but I believe in How-To guide we should really be concise and straight to the point, No explanation required (we could start putting general concepts in the readme, but not pollute the How-To guide, and have the latter as clean as possible).

deusebio (approved):
All good!

welpaolo (approved):

PR #11, opened 2022-11-22
DPE-855: Adding more unit tests using mock framework to improve test coverage 

averma-canonical:


tests/unittest/test_helpers.py
```diff
@@ -166,6 +169,463 @@ def test_merge_configurations(self):
...
+
+        cmd_get_master = f"{test_id}/kubectl --kubeconfig {kubeconfig} --namespace {namespace} --context {context} config view --minify -o jsonpath=\"{{.clusters[0]['cluster.server']}}\""
+        output_get_master = control_plane_uri.encode("utf-8")
+
+        values = {
+            cmd_get_service_account: output_get_service_account,
+            cmd_get_master: output_get_master,
+        }
+
+        def side_effect(*args, **kwargs):
```
deusebio, line 227: Move this at the top of the tests (line 201 or so)

deusebio:
90% code coverage where we have business logic is very very very very good!

tests/unittest/test_helpers.py
```diff
@@ -166,6 +173,981 @@ def test_merge_configurations(self):
...
+        mock_sys_exit.return_value = 0
+
+        conf_list = [
+            "spark.kubernetes.namespace",
+            "spark.kubernetes.authenticate.driver.serviceAccountName",
+        ]
+
+        overrides = helpers.utils.parse_conf_overrides(conf_list)
+
+        mock_sys_exit.assert_any_call(helpers.constants.EXIT_CODE_BAD_CONF_ARG)
```
deusebio, line 215: Very very good to tests happy path but most importantly when things go wrong!!!!!

:+1:

tests/unittest/test_helpers.py
```diff
@@ -166,6 +173,981 @@ def test_merge_configurations(self):
...
+
+        mock_subprocess.side_effect = side_effect_subprocess
+
+        test_id = str(uuid.uuid4())
+        username = str(uuid.uuid4())
+        namespace = str(uuid.uuid4())
+        kubeconfig = str(uuid.uuid4())
+        context = str(uuid.uuid4())
+        conf_key = str(uuid.uuid4())
+        conf_value = str(uuid.uuid4())
```
deusebio, line 469: Just saying but it does not need refactoring, giving that we will anyway need to refactor unittests, after the structure refactoring we are planning, but many tests have very similar piece of code for mocking. This could in general be factor out 

But as I said, just raising this point when we will restructure unit tests next time. At this point, there is no action required!

tests/unittest/test_helpers.py
```diff
@@ -166,6 +173,981 @@ def test_merge_configurations(self):
...
+        context1 = str(uuid.uuid4())
+        token1 = str(uuid.uuid4())
+        username2 = str(uuid.uuid4())
+        context2 = str(uuid.uuid4())
+        token2 = str(uuid.uuid4())
+        username3 = str(uuid.uuid4())
+        context3 = str(uuid.uuid4())
+        token3 = str(uuid.uuid4())
+
+        kubeconfig_yaml = {
```
deusebio, line 781: If may be best to store a YAML file in the tests folder and try to read from there. 

Also in this case, this does not need to be addressed now, but just keep in mind for the next tests restructuring

deusebio (approved):
Very good! 90% coverage! I approve the PR!!!

I have some points worth bringing those up not to be addressed in this PR, but rather for next restructuring.

tests/unittest/test_helpers.py
welpaolo: I also suggest to have to have a module that contains shared functions for the tests.

welpaolo (approved):
The tests looks good, I will suggest to have a docstring description for each test descriptioning its purpose. No need to do in this PR

PR #12, opened 2022-11-23
DPE-840: Introducing pyspark wrapper.

averma-canonical:
Introducing wrapper for pyspark introduced issues.

1. Mixed mode exposes python version mismatch between snap (3.8) and rock (3.10) coming from core20 and core22 resp.
2. Moving the snap to core22 is not a trivial change
3. Found this bug, configuration "associated" with a service account can by mistake point to another service account which creates havoc during job submission and spark-shell and pyspark runs.

### DIFF
--- a/tests/integration/ie-tests.sh
+++ b/tests/integration/ie-tests.sh
@@ -84,7 +84,7 @@ test_pyspark() {
   echo "     return 1 if x ** 2 + y ** 2 <= 1 else 0" >> test-pyspark.py
   echo "count = spark.sparkContext.parallelize(range(1, n + 1), partitions).map(f).reduce(add)" >> test-pyspark.py
   echo "print (\"Pi is roughly %f\" % (4.0 * count / n))" >> test-pyspark.py
-  echo -e "$(cat test-pyspark.py | spark-client.pyspark --conf "spark.driver.host=${DRIVER_IP} --conf spark.executor.instances=4")" > pyspark.out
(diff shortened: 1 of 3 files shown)


tests/integration/ie-tests.sh
```diff
@@ -63,8 +63,41 @@ test_spark_shell() {
...
+  echo "spark = SparkSession(sc)" >> test-pyspark.py
+  echo "for conf in spark.sparkContext.getConf().getAll(): print (conf)" >> test-pyspark.py
+  echo "partitions = 1000" >> test-pyspark.py
+  echo "n = 100000 * partitions" >> test-pyspark.py
+  echo "def f(_: int) -> float:" >> test-pyspark.py
+  echo "     x = random() * 2 - 1" >> test-pyspark.py
+  echo "     y = random() * 2 - 1" >> test-pyspark.py
+  echo "     return 1 if x ** 2 + y ** 2 <= 1 else 0" >> test-pyspark.py
+  echo "count = spark.sparkContext.parallelize(range(1, n + 1), partitions).map(f).reduce(add)" >> test-pyspark.py
+  echo "print (\"Pi is roughly %f\" % (4.0 * count / n))" >> test-pyspark.py
```
deusebio, line 86:
```suggestion
  echo "print (\"Pi is roughly %f\" % round(4.0 * count / n, 2))" >> test-pyspark.py
```

deusebio (approved):
All good! Just a small note to make the process more robust. I'm afraid sometimes Python does truncation more than rounding. This should make things more robust on the tests.

welpaolo (approved):
Looks good to me

PR #16, opened 2022-12-08
DPE-1020: Added unit tests for domain layer.

averma-canonical:


tests/unittest/test_services.py
```diff
@@ -18,6 +28,865 @@ def test_conf_expansion_cli(self):
...
+        username2 = str(uuid.uuid4())
+        context2 = str(uuid.uuid4())
+        token2 = str(uuid.uuid4())
+        username3 = str(uuid.uuid4())
+        context3 = str(uuid.uuid4())
+        token3 = str(uuid.uuid4())
+        test_kubectl_cmd = str(uuid.uuid4())
+
+        kubeconfig_yaml = {
+            "apiVersion": "v1",
```
deusebio, line 47: it would be best to put the yaml as a YAML file, and then read it. 

We could place the YAML file in tests/resource/ or tests/unittest/resource, either is fine by me. 

See these links for some inspiration:

https://github.com/CGnal/cgnal-core/blob/main/tests/__init__.py#L9
https://github.com/CGnal/cgnal-core/blob/main/tests/test_config.py#L55
https://github.com/CGnal/cgnal-core/tree/main/tests/resources/data

spark_client/services.py
```diff
@@ -654,7 +660,7 @@ def set_primary(self, account_id: str) -> str:
 
         if any([account.primary for account in self.all()]):
             self.logger.info("Switching primary account")
-            for account_id, account in self.cache.items():
+            for account_id_i, account in self.cache.items():
```
deusebio, line 663: If not used, we could also just use the placeholder "_" (it is sort of a convetion) or rather do "self.cache.values()" instead

deusebio (commented):
Just some comments. The only relevant one is to try to move to a central point the YAML files. I would also just put some values (with strange names or ids) without having to make them random all the time. I would prefer centralization of YAML and simpler code in the python test code, over fully random parameters on YAML that however requires us to have a lot of YAML code in the python tests files

Wondering whether @welpaolo agrees or if he has any other preference on this?

tests/unittest/test_services.py
deusebio: Let's leave this out for another refactoring or improvemnt. 

This is a minor point

averma-canonical: Thanks Enrico. Addressed rest of the comments.

spark_client/services.py
```diff
@@ -275,6 +275,11 @@ def select_by_master(self, master: str):
         }
 
         self.logger.debug(f"Clusters API: {dict(api_servers_clusters)}")
+        for _context in self.kube_config["contexts"]:
+            print(_context["name"])
```
deusebio, line 279: Why do we still have prints here?

averma-canonical: oops, this was when I was writing down the tests to check what to assert for. sorry. let me fix these.

deusebio (approved):
Good to go!

PR #17, opened 2022-12-14
DPE-1022: Illustrating RBAC addition to service account role.

averma-canonical:
Hi @deusebio, sharing this to illustrate the skeleton of changes needed to implement rbac. We are of course waiting for the stakeholders to agree on final set of permissions to be used.

### DIFF
--- a/spark_client/services.py
+++ b/spark_client/services.py
@@ -226,7 +226,9 @@ class KubeInterface(WithLogging):
                         --role view in the command
         """
 
-        formatted_extra_args = " ".join([f"--{k}={v}" for k, values in extra_args.items() for v in values])
+        formatted_extra_args = " ".join(
+            [f"--{k}={v}" for k, values in extra_args.items() for v in values]
+        )
         self.exec(
             f"create {resource_type} {resource_name} {formatted_extra_args}",
             namespace=namespace,
@@ -501,10 +503,7 @@ class K8sServiceAccountRegistry(AbstractServiceAccountRegistry):
             "role",
             rolename,
             namespace=service_account.namespace,
-            **{
-                "resource": ["pods", "configmaps"],
-                "verb": ["create", "get", "watch"]
-            },
+            **{"resource": ["pods", "configmaps"], "verb": ["create", "get", "watch"]},
         )
         self.kube_interface.create(
             "rolebinding",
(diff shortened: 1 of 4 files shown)


spark_client/services.py
```diff
@@ -226,7 +226,9 @@ def create(
                         --role view in the command
         """
 
-        formatted_extra_args = " ".join([f"--{k}={v}" for k, v in extra_args.items()])
+        formatted_extra_args = " ".join(
+            [f"--{k}={v}" for k, values in extra_args.items() for v in values]
```
deusebio, line 230: I would probably allow people to also feed normal values, rather than lists (e.g. documentation should be updated otherwise). But what about we do:

```
def listify(value: Any) -> List[str]:
    return [str(v) for v in values] if isinstace(value, list) else [str(value)] 
[f"--{k}={v}" for k, values in extra_args.items() for v in listify(values)]
```

We should anyway update doc string to clarify that lists will be treated

spark_client/services.py
```diff
@@ -478,7 +480,7 @@ def set_primary(self, account_id: str) -> str:
         )
         self.kube_interface.set_label(
             "rolebinding",
-            f"{service_account.name}-role",
+            f"{service_account.name}-role-binding",
```
deusebio, line 483: :+1:

spark_client/services.py
averma-canonical: Sure, this helps retain the old interface as well which was nicer than having to pass everything as a list. Thanks Enrico.

spark_client/services.py
```diff
@@ -491,18 +493,26 @@ def create(self, service_account: ServiceAccount) -> str:
...
 
         self.kube_interface.create(
             "serviceaccount", service_account.name, namespace=service_account.namespace
         )
+        self.kube_interface.create(
+            "role",
+            rolename,
+            namespace=service_account.namespace,
+            **{
+                "resource": ["pods", "pods/log", "configmaps", "services"],
```
deusebio, line 507: I would skip "pods/log" resource following the MM discussion

averma-canonical:
Hi @delgod  and @deusebio,

Tested rbac with non-default namespace. Works fine!

deusebio (approved):
Looks good!

PR #21, opened 2023-01-16
Removing incorrect condition for snap publish....

averma-canonical:
… Learnt that multiple python version runs have no bearing on single publish of snap.


deusebio (approved):
Good to go

welpaolo (approved):
Looks good

PR #23, opened 2023-01-17
Draft to discuss github release for spark-client.

averma-canonical:


.github/workflows/release.yaml
```diff
@@ -0,0 +1,75 @@
...
+          python-version: ${{matrix.python-version}}
+          architecture: x64
+      - id: code-checks
+        name: Make sure the code follows conventions and tests correctly
+        run: |
+          make checks
+      - id: build-package
+        name: Build package
+        run: |
+          poetry build
```
averma-canonical, line 40: assuming this is different from "make build" that we do for the snap.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,75 @@
+name: Create a release
+
+env:
+  BRANCH: release
+  VERSION: 3.3.1
```
averma-canonical, line 5: How to appropriately regex the BRANCH here?

How to make the VERSION come dynamically?

.github/workflows/release.yaml
deusebio: yes, it is different. In the SNAP, with build, we build the snap, not the package

.github/workflows/release.yaml
```diff
@@ -0,0 +1,75 @@
...
+          poetry build
+      - name: Add version to environment vars
+        run: |
+          echo "PROJECT_VERSION=${{ env.VERSION }}" >> $GITHUB_ENV
+      - name: Check if tag version matches project version
+        run: |
+          TAG=$(git describe HEAD --tags --abbrev=0)
+          echo $TAG
+          echo $PROJECT_VERSION
+          if [[ "$TAG" != "v$PROJECT_VERSION" ]]; then exit 1; fi
```
averma-canonical, line 49: Branch tagging would be a manual process I believe.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,75 @@
...
+  test:
+    name: Test Release
+    runs-on: ubuntu-latest
+    needs:
+      - build
+    steps:
+      - name: Download tar.gz file
+        run: |
+          mkdir pkgs
+          # curl download the release and test ?
```
averma-canonical, line 74: Is post publish test useful?

.github/workflows/release.yaml
averma-canonical: :+1:

.github/workflows/release.yaml
deusebio: I would say so!

averma-canonical:
This was useful to discuss and redesign the publish workflow.

PR #28, opened 2023-01-30
DPE-1152 - Refactor parsing to introduce flexibility in options order!

averma-canonical:
DPE-1152 - Factor out all snap environment related hard-coding in snap client (SNAP, SNAP_USER_DATA etc.)
DPE-1150 - Undo renaming of spark-submit and spark-shell entry points in snap
DPE-1148 - Client UX : order of options should not be a restriction, currently options need to be specified before verb but primary goes after the verb

### DIFF
--- a/spark_client/domain.py
+++ b/spark_client/domain.py
@@ -157,14 +157,14 @@ class Defaults:
         self.environ = environ if environ is not None else {}
 
     @property
-    def snap_folder(self) -> str:
+    def spark_folder(self) -> str:
         """Return the SNAP folder"""
         return self.environ.get("SNAP", self.environ.get("HOME"))
 
     @property
     def static_conf_file(self) -> str:
         """Return static config properties file packaged with the client snap."""
-        return f"{self.snap_folder}/conf/spark-defaults.conf"
+        return f"{self.spark_folder}/conf/spark-defaults.conf"
 
     @property
     def dynamic_conf_file(self) -> str:
@@ -201,7 +201,7 @@ class Defaults:
     @property
     def kubectl_cmd(self) -> str:
         """Return default kubectl command."""
-        return f"{self.snap_folder}/kubectl" if self.snap_folder else "kubectl"
+        return f"{self.spark_folder}/kubectl" if self.spark_folder else "kubectl"
 
     @property
     def scala_history_file(self):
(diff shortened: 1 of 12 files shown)


spark_client/domain.py
```diff
@@ -159,22 +159,22 @@ def __init__(self, environ: Dict = dict(os.environ)):
     @property
     def snap_folder(self) -> str:
```
deusebio, line 160: maybe we could name this a bit more general than snap, since the CLI will work also out of snap. Maybe we could call this "spark_folder" instead

deusebio (approved):
Just a rather minor comment, that however I believe it would improve readability of the code

deusebio (dismissed):
Good to go!

taurus-forever (dismissed):
LGTM

spark_client/domain.py
```diff
@@ -157,24 +157,24 @@ def __init__(self, environ: Dict = dict(os.environ)):
         self.environ = environ if environ is not None else {}
 
     @property
-    def snap_folder(self) -> str:
+    def spark_folder(self) -> str:
         """Return the SNAP folder"""
-        return self.environ["SNAP"]
+        return self.environ.get("SNAP", self.environ.get("HOME"))
```
deusebio, line 162: Sorry, for not realizing this earlier, but it is somewhat connected to the other PR. Here we should probably decide where this env variable should be "SPARK_HOME"

spark_client/domain.py
```diff
@@ -157,24 +157,24 @@ def __init__(self, environ: Dict = dict(os.environ)):
...
     def static_conf_file(self) -> str:
         """Return static config properties file packaged with the client snap."""
-        return f"{self.environ.get('SNAP')}/conf/spark-defaults.conf"
+        return f"{self.spark_folder}/conf/spark-defaults.conf"
 
     @property
     def dynamic_conf_file(self) -> str:
         """Return dynamic config properties file generated during client setup."""
-        return f"{self.environ.get('SNAP_USER_DATA')}/spark-defaults.conf"
+        return f"{self.environ.get('SNAP_USER_DATA', self.environ.get('SPARK_HOME'))}/spark-defaults.conf"
```
deusebio, line 172: This should probably be the USER home in outside of snaps

deusebio (approved):
Looks very good! Only one small comment

spark_client/domain.py
averma-canonical: Thanks Enrico, made the change.

spark_client/domain.py
averma-canonical: pinned to SPARK_HOME in non-snap settings.

taurus-forever (approved):
Nice!

deusebio (approved):
LGTM

PR #31, opened 2023-02-15
[MISC] CLI argument parsing restructuring

deusebio:
This is to fix the user-experience that was basically ignoring kubeconfig inputs from the CLI 

Meanwhile I refactored a bit the argument parser logic trying to factor out common parser. We should make sure that the CLI is still working as expected. I'm relying on integration tests, I'm just wondering whether they are enough.


spark_client/utils.py
```diff
@@ -208,45 +221,74 @@ def add_logging_arguments(parser):
...
+        action="append",
         type=str,
-        help="Namespace of service account name to use other than primary.",
+        help="Config properties to be added to the service account.",
     )
-
     return parser
 
 
-def add_deploy_arguments(parser):
```
averma-canonical, line 249: can we please make default as 'cluster' instead of 'client' currently.

tests/unittest/test_argument_parsing.py
```diff
@@ -0,0 +1,175 @@
...
+
+        self.assertRaises(
+            argparse.ArgumentError,
+            lambda: self.parser.parse_known_args(["--log-level", "NON-EXISTING"]),
+        )
+
+    def test_spark_defaults(self):
+        args, extra_args = self.parser.parse_known_args()
+        self.assertEqual(args.namespace, "default")
+        self.assertEqual(args.username, "spark")
```
averma-canonical, line 43: deploy-mode cluster please. makes the base use case succinct.

welpaolo (approved):
LGTM

averma-canonical (approved):
Great

PR #32, opened 2023-02-17
Fixing CI token for PR on Spark update detection and for Github release publish.

averma-canonical:
New release of Spark is out (3.3.2). Our CI pipeline detected it but failed to create a PR for this due to bad token configuration.

With suggestions from Alex, sharing a fix for broken github token in CI for
- PR creation on Spark update detection
- Github artifacts release pipeline

Also, replacing third party action with github tooling for PR creation.

Adding Carl to the review as well for suggestions.

### DIFF
--- a/.github/workflows/on_spark_update_available.yaml
+++ b/.github/workflows/on_spark_update_available.yaml
@@ -69,8 +69,8 @@ jobs:
           --title 'Update snap for new spark version' \
           --body 'Update client snap with new version of Apache Spark.' \
           --base main
-          --assignee team-reviewers, deusebio, taurus, abhishek-verma \
-          --reviewer team-reviewers, deusebio, taurus, abhishek-verma \
+          --assignee team-reviewers, deusebio, taurus-forever, averma-canonical \
+          --reviewer team-reviewers, deusebio, taurus-forever, averma-canonical \
           --label 'automated pr'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


.github/workflows/on_spark_update_available.yaml
```diff
@@ -65,23 +65,12 @@ jobs:
...
-            owners
-            maintainers
-          draft: false
\ No newline at end of file
+        run: gh pr create \
+          --title 'Update snap for new spark version' \
+          --body 'Update client snap with new version of Apache Spark.' \
+          --base main
+          --assignee team-reviewers, deusebio, taurus, abhishek-verma \
+          --reviewer team-reviewers, deusebio, taurus \
```
taurus-forever, line 73: why don't you want to review? :-D

taurus-forever (commented):
JFYI `gh pr create` was not tested on my side, can you please ensure it does the job?
You can use https://github.com/lhotari/action-upterm to enter SSH into the runner and test all commands locally, like a runner.

deusebio:
> Looks good!
> 
> FYI the GITHUB_TOKEN being used in a job with third-party actions may have security risks
> 
> Also the third-party action to create a release can be replaced with the GitHub CLI. See here for an example: https://github.com/canonical/data-platform-workflows/pull/9/files

This action was also referred in this [github page](https://github.com/actions/create-release), so sounded sort of reliable, but I'd be more than happy to change it if we have other suggestions or tools that we recommend to use as data platform, to be standard across repos.

deusebio (approved):
Looks good to me! If it is not too much work, I'd would follow up on @carlcsaposs-canonical comment. But if it is not straight forward, we can also approach that in a separate ticket

taurus-forever (approved):
LGTM, once again, I had no chance to test `gh pr create` in full on my side, this is why https://github.com/canonical/mysql-k8s-bundle is still open.

PR #34, opened 2023-02-17
[CRITICAL] Runners blocked. Removing broken integration tests, Spark version mismatch.

averma-canonical:
Spark version has moved to 3.3.2. This causes java/scala version mismtach between built snap and the rock image.

Result, integration tests for shell and pyspark get stuck infinitely engaging the runners.

Removing the tests until this problem is fixed with a proper version management design.

### DIFF
--- a/.github/workflows/release_github.yaml
+++ b/.github/workflows/release_github.yaml
@@ -2,7 +2,7 @@ name: Create a github release
 
 env:
   BRANCH: ${{ github.ref_name }}
-  SNAP_VERSION: 3.3.1
+  SNAP_VERSION: 3.3.2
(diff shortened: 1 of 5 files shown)


averma-canonical:
Hi @deusebio and @taurus-forever,

Apologies but I would request again to disable the shell and pyspark integration tests for now till we get these things ironed out properly.

In snap static configuration, there are references to spark rock 3.3.1 which need to be updated. 

Automated version management needs to be thought through properly before we enable such sensitive tests.

deusebio (approved):
Looks good!

taurus-forever (approved):

PR #36, opened 2023-02-21
1035 driver host autodetect

averma-canonical:
Currently spark.driver.host config option has to be supplied explicitly by the user. 

This change is to choose the IP for the user when we are reasonably sure.
i.e. if output of socket.gethostbyname(socket.gethostname()) in python and hostname -I first entry point to the same ip.

### DIFF
--- a/spark_client/services.py
+++ b/spark_client/services.py
@@ -838,7 +838,9 @@ class SparkInterface(WithLogging):
                 "spark.driver.host" not in conf.props
                 and not self.check_driver_host_extra_conf()
             ):
-                raise ValueError("")
+                raise ValueError(
+                    "Please specify spark.driver.host configuration property"
+                )
 
             conf.write(t.file)
             t.flush()
--- a/spark_client/services.py
+++ b/spark_client/services.py
@@ -817,13 +817,8 @@ class SparkInterface(WithLogging):
         ) as t:
             self.logger.debug(f"Spark props available for reference at {t.name}\n")
 
-            spark_driver_host = self.detect_host()
-
             conf = (
-                PropertyFile({"spark.driver.host": spark_driver_host})
-                if spark_driver_host
-                else PropertyFile({})
-                + self._read_properties_file(self.defaults.static_conf_file)
+                self._read_properties_file(self.defaults.static_conf_file)
                 + PropertyFile(
                     {
                         "spark.driver.extraJavaOptions": f"-Dscala.shell.histfile={self.defaults.scala_history_file}"
@@ -833,6 +828,8 @@ class SparkInterface(WithLogging):
                 + self._read_properties_file(self.defaults.env_conf_file)
                 + self._read_properties_file(cli_property)
             )
+
+            conf = self.add_optional_detected_driver_host(conf)
 
             if (
                 "spark.driver.host" not in conf.props
@@ -870,17 +867,14 @@ class SparkInterface(WithLogging):
         ) as t:
             self.logger.debug(f"Spark props available for reference at {t.name}\n")
 
-            spark_driver_host = self.detect_host()
-
             conf = (
-                PropertyFile({"spark.driver.host": spark_driver_host})
-                if spark_driver_host
-                else PropertyFile({})
-                + self._read_properties_file(self.defaults.static_conf_file)
+                self._read_properties_file(self.defaults.static_conf_file)
                 + self.service_account.configurations
                 + self._read_properties_file(self.defaults.env_conf_file)
                 + self._read_properties_file(cli_property)
             )
+
+            conf = self.add_optional_detected_driver_host(conf)
 
             if (
                 "spark.driver.host" not in conf.props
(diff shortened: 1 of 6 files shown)


grobbie:
Is it not possible to set `spark.driver.bindAddress=0.0.0.0` and then set `spark.driver.host` to the value of `hostname -f` instead of doing this? Granted that will only work when the host has a hosttname that is resolvable by the executors. But at first glance this PR seems quite complex and potentially error prone to resolve an issue that should essentially be "resolved" via DNS.

spark_client/utils.py
```diff
@@ -259,3 +274,49 @@ def add_deploy_arguments(parser):
...
+def detect_host() -> Any:
+    try:
+        host_py = socket.gethostbyname(socket.gethostname()).strip()
+    except Exception:
+        return None
+
+    try:
+        host_bash = (
+            subprocess.check_output(
+                "hostname -I | cut -d' ' -f1", shell=True, stderr=None
```
averma-canonical, line 311: Hi @grobbie - this is where we are detecting the host. By 2 methods, which if give same ip, we use it.

grobbie: sure, but `hostname -f` will give you the fully qualified domain name of the computer. The executors would then use DNS to find the correct IP of the computer hosting the Spark driver. And if DNS doesn't work properly, there are going to be bigger problems. I don't think we should try to guess the IP, especially if the computer is running Docker or LXD, or his multihomed with several NICs (eg. fitted with iLO and public/private IPs).

averma-canonical: Thanks Rob, that makes sense. Let me verify this.

averma-canonical: getting ```java.net.UnknownHostException``` in executors when passing the ```hostname -f``` output.

spark_client/utils.py
```diff
@@ -259,3 +274,49 @@ def add_deploy_arguments(parser):
         help="Deployment mode for job submission. Default is 'client'.",
     )
     return parser
+
+
+def get_driver_host_conf() -> Any:
```
deusebio, line 279: **thought** I've been thinking a bit about this. I'm starting to think that the best place for this would be the SparkInterface, when merging configurations. If we have a valid default (because of the auto-inference) we place that in the merged properties file, before reading the env property file and the one provided by the user. 

We then raise an error if the spark.driver.host was not in the merged properties OR in the extra_confs.

spark_client/utils.py
deusebio: I mean, the idea here is not to work in 100% cases, but to work in most of them, and get in most of the cases a cleaner UX (like showed in the tests, [here](https://github.com/canonical/spark-client-snap/pull/36/files#diff-c841d3a6788870a95eca90d4ff1523ea3a3f3a333f7b883b0e78f7b68cd19526R72)). When the auto-inference does not work, the user can of course specify the driver.host at different level of out hierarchical configuratoin system:
* in the env file (which should be static in the machine, there may be a sensible place to specify machine IP)
* in the property file provided via CLI
* as --conf in the CLI

Of course, the logic behind auto-inference can be changed and improved as we go, but this PR should nevertheless pose the structure to allow this

deusebio (commented):
I only have a relevant comment that I believe this logic should rather be in the SparkInterface such that we can easily integrate this in the hierarchical configuration system we implemented

spark_client/services.py
```diff
@@ -870,3 +897,46 @@ def pyspark_shell(self, cli_property: Optional[str], extra_args: List[str]):
             self.logger.debug(submit_cmd)
             with environ(KUBECONFIG=self.kube_interface.kube_config_file):
                 os.system(submit_cmd)
+
+    def add_optional_detected_driver_host(self, conf: PropertyFile):
```
welpaolo, line 901: Can you please add a documentation for this function and the other added functions

welpaolo (approved):
LGTM

spark_client/services.py
```diff
@@ -870,3 +897,46 @@ def pyspark_shell(self, cli_property: Optional[str], extra_args: List[str]):
...
+            return None
+
+        try:
+            host_bash = (
+                subprocess.check_output(
+                    "hostname -I | cut -d' ' -f1", shell=True, stderr=None
+                )
+                .decode("utf-8")
+                .strip()
+            )
```
delgod, line 935: I think logic can be more reliably (idea taken here - https://stackoverflow.com/a/166589)
* we have IP and port of kubernetes API in `self.service_account.api_server` var.
* spark client must be able to connect to it
* if we can connect to it, we can return our IP from TCP connection
```
import socket
host = self.service_account.api_server.split(":")[1].split("/")[-1]
port = self.service_account.api_server.split(":")[2] if len(api_server.split(":")) == 3 else 433

s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
s.connect((host, port))
print(s.getsockname()[0])
s.close()
```

delgod (changes requested):

spark_client/services.py
deusebio: I really like this idea!

deusebio:
should we close this PR since it is superseded by #38 ?

averma-canonical:
> should we close this PR since it is superseded by #38 ?

Correct. The new rebased PR takes care of this now.

PR #37, opened 2023-02-22
Adding 2 hr timeout for integration tests which can be unstable as Spark versions change.

averma-canonical:
…as requested in DPE-1326

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,6 +64,7 @@ jobs:
 
   test:
     name: Test Snap
+    timeout-minutes: 120
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


deusebio (changes requested):
Probably we could even set this to 20minutes. 

It everything goes fine it really takes less than 10min. If we are over 20min, there is probably something wrong so we could kill this. If in the future we will put a lot of itegration tests and we will start hitting the limit to increase this, we will of course free to do this. 

But, as of today, I'll just roughly double what it the actual time

taurus-forever (approved):
20 mins is even better.

deusebio (approved):

PR #44, opened 2023-03-20
Introducing lighkube, draft for discussion.

averma-canonical:
Open Issues
- lint problems (mypy)


spark_client/services.py
```diff
@@ -112,6 +95,423 @@ def user(self):
...
+            else self.with_context(contexts_for_api_server[0])
+        )
+
+
+class LightKube(AbstractKubeInterface):
+    def __init__(
+        self,
+        kube_config_file: Union[str, Dict[str, Any]],
+        context_name: Optional[str] = None,
+        kubectl_cmd="kubectl",
```
deusebio, line 209: **todo** We don't need this anymore, do we?

deusebio (commented):
Very nice! This is exactly what I had in mind, but i was not sure it was really possible to have two implementations: one with kubectl and the other one with lightkube, that one can use interchangeably in the registry!!!

Well done!!!

I just had a quick glance and everything seems to me to go in the right direction, but I just wanted to play a bit around with this.

spark_client/services.py
```diff
@@ -112,6 +96,482 @@ def user(self):
...
+                        will be prepended with the -- in the cmd, e.g. {"role": "view"} will translate as
+                        --role=view in the command. List of parameter values against a parameter key are also accepted.
+                        e.g. {"resource" : ["pods", "configmaps"]} which would translate to something like
+                        --resource=pods --resource=configmaps
+        """
+
+        res = None
+        if resource_type == "serviceaccount":
+            with open(
+                f"{os.getenv('SNAP')}/python/spark_client/resources/lightkube_jinja_templates/serviceaccount_yaml.tmpl"
```
deusebio, line 449: **issue** this won't work outside of snap. Beside, in general we should never access envs using os.getenv. We defined the Defaults object to be able to proxy and centralize the envs that we use, such that we have control on these. If we use os.env scattered around, we then lose this control 

Anyhow, these are resources and as such, there is a pythonic way to access them which is using this library

https://importlib-resources.readthedocs.io/en/latest/using.html

spark_client/services.py
```diff
@@ -112,6 +96,482 @@ def user(self):
...
+            resource_type: type of the resource to be labeled, e.g. service account, rolebindings, etc.
+            resource_name: name of the resource to be labeled
+            label: label to remove
+            namespace: namespace where the resource is
+        """
+        label_to_remove = f"/metadata/labels/{label.replace('/','~1')}"
+        self.logger.debug(f"Removing label {label_to_remove}")
+        patch = [{"op": "remove", "path": label_to_remove}]
+
+        if resource_type == "serviceaccount":
```
deusebio, line 389: **suggestion** We could use Enum for handling resource_type, given they are values chosen from a set of values

https://docs.python.org/3/library/enum.html

This should make it more typed, and also avoid to use strings, which are error prone

deusebio (changes requested):
Code is shaping up well! I really enjoyed how the code had been structured. There a few issues that needs to be addressed before merging though

spark_client/services.py
```diff
@@ -464,19 +937,22 @@ def set_primary(self, account_id: str) -> str:
         """
 
         # Relabeling primary
-        primary_account = self.get_primary()
+        try:
+            primary_account = self.get_primary()
```
deusebio, line 941: **todo** I believe this does not comply with the signature. The method is supposed to return None if there is no primary. There was a bug in the code, that I have corrected in the PR #45

spark_client/cli/service_account_registry.py
```diff
@@ -126,18 +131,47 @@ def create_service_account_registry_parser(parser: ArgumentParser):
         print(user_id)
```
welpaolo, line 131: logger

welpaolo (approved):
Looks good to me. Please resolve the comments done by Enrico

tests/unittest/resources/data/kubeconfig.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: v1
```
grobbie, line 1: I don't like this getting committed to git's memory full of test public keys and tokens. Please remove this file from the PR and generate one during test setup instead.

grobbie (changes requested):
Please remove the test kubeconfig file from this PR and generate one in a test setup method instead

deusebio (approved):
LGTM

grobbie (approved):
thanks

PR #46, opened 2023-03-23
Pod based integration tests.

averma-canonical:
Issues:
- [DPE-1545](https://warthogs.atlassian.net/browse/DPE-1545) with restricted account based test with ```spark-submit```.

### DIFF
--- a/tests/integration/ie-tests.sh
+++ b/tests/integration/ie-tests.sh
@@ -174,6 +174,157 @@ cleanup_user_failure() {
   cleanup_user 1 spark tests
 }
 
+setup_test_pod() {
+  kubectl apply -f ./tests/integration/resources/testpod.yaml
+
+  for i in {1..5}
+  do
+    pod_status=$(kubectl get pod testpod | awk '{ print $3 }' | tail -n 1)
+    if [ "${pod_status}" == "Running" ]
+    then
+        echo "testpod is Running now!"
+        break
+    else
+        sleep 5
+    fi
+  done
+
+  MY_KUBE_CONFIG=$(cat /home/${USER}/.kube/config)
+  MY_SPARK_CONFIG=$(cat /home/${USER}/conf/spark-defaults.conf)
+
+  kubectl exec -it testpod -- /bin/bash -c 'mkdir /home/spark/.kube'
+  kubectl exec -it testpod -- env KCONFIG="$MY_KUBE_CONFIG" /bin/bash -c 'echo "$KCONFIG" > /home/spark/.kube/config'
+  kubectl exec -it testpod -- /bin/bash -c 'cat /home/spark/.kube/config'
+
+  kubectl exec -it testpod -- /bin/bash -c 'mkdir /home/spark/conf'
+  kubectl exec -it testpod -- env SCONFIG="$MY_SPARK_CONFIG" /bin/bash -c 'echo "$SCONFIG" > /home/spark/conf/spark-defaults.conf'
+  kubectl exec -it testpod -- /bin/bash -c 'cat /home/spark/conf/spark-defaults.conf'
+}
+
+teardown_test_pod() {
+  kubectl delete pod testpod
+}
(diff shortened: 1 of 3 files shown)


tests/integration/resources/testpod.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: testpod
+spec:
+  containers:
+  - image: dataplatformoci/spark:3.3.2@sha256:5d5b02b7fcc60f86fa07c9aefbe98245d5030e10be12af9968b8b9b545a97355
```
averma-canonical, line 7: @delgod and @deusebio 

I had to explicitly add the latest SHA256 id here to work with the latest ROCK image to create the test pod.

Can you please share if this can be removed for CI where runners do not have any cached previous version?

deusebio: uhm, you mean that otherwise it does work locally or in CI? I would say that in CI, since there is no image downloaded yet, k8s would pull the latest image on dockerhub, which should already have the right hash.

Am I missing something?

tests/integration/ie-tests.sh
```diff
@@ -174,6 +174,157 @@ cleanup_user_failure() {
...
+  run_spark_shell_in_pod tests spark
+}
+
+run_pyspark_in_pod() {
+  echo "run_pyspark_in_pod ${1} ${2}"
+
+  NAMESPACE=$1
+  USERNAME=$2
+
+  echo "import sys" > test-pyspark.py
```
welpaolo, line 283: Is it possible to create a *.py file and  put it into the test folder, instead of doing it with the following commands ?

welpaolo (approved):
LGTM, I left a non blocking comment

tests/integration/ie-tests.sh
```diff
@@ -174,6 +174,157 @@ cleanup_user_failure() {
...
+  echo "     return x * x + y * y < 1 " >> test-pyspark.py
+  echo "count = spark.sparkContext.parallelize(range(n), partitions).filter(f).count()" >> test-pyspark.py
+  echo "print (\"Pi is roughly %f\" % (4.0 * count / n))" >> test-pyspark.py
+
+  PYSPARK_COMMANDS=$(cat ./test-pyspark.py)
+
+  echo -e "$(kubectl exec testpod -- env UU="$USERNAME" NN="$NAMESPACE" CMDS="$PYSPARK_COMMANDS" /bin/bash -c 'echo "$CMDS" | spark-client.pyspark --username $UU --namespace $NN')" > pyspark.out
+
+  cat pyspark.out
+  pi=$(cat pyspark.out  | grep "Pi is roughly" | tail -n 1 | rev | cut -d' ' -f1 | rev | cut -c 1-3)
```
taurus-forever, line 304: Can you please share content of pyspark.out.
What happening here is a madness. 
1) `cat file | grep match`  => `grep match file`
2) two times `rev` and two times `cut` ? o_0

tests/integration/ie-tests.sh
```diff
@@ -174,6 +174,157 @@ cleanup_user_failure() {
...
+  echo -e "Pyspark Pi Job Output: \n ${pi}"
+  rm test-pyspark.py pyspark.out
+  if [ "${pi}" != "3.1" ]; then
+      exit 1
```
taurus-forever, line 308: Maybe some error message for aborting the test?
Newbies do not know the magic value `3.1` :-)

taurus-forever (changes requested):
Shell file can be significantly updated, up to Enrico here.
Please switch to ghci.io registry.

deusebio (approved):
LGTM

deusebio:
> LGTM. P.S. it is a way better now! Just a question, why python files saved as txt?

Yes, maybe just for the sake of better readability, we could rename them as ".py" and ".scala"

PR #57, opened 2023-04-26
DPE-1609 updating spark files arrangement in snap as per the spec

averma-canonical:
Updating snap organization scheme as per the snap and rock packaging specification - https://docs.google.com/document/d/1MyT4Wt_-vgPwhwEjC-oeHRdLYrZWaQzFGTwclT7hD1U/edit

### DIFF
--- a/helpers/utils.py
+++ b/helpers/utils.py
@@ -173,7 +173,7 @@ def merge_configurations(dictionaries_to_merge: List[Dict]) -> Dict:
 
 def get_static_defaults_conf_file() -> str:
     """Return static config properties file packaged with the client snap."""
-    SPARK_STATIC_DEFAULTS_FILE = f"{os.environ.get('SNAP')}/conf/spark-defaults.conf"
+    SPARK_STATIC_DEFAULTS_FILE = f"{os.environ.get('SNAP_DATA')}/etc/spark/spark-defaults.conf"
     return SPARK_STATIC_DEFAULTS_FILE
 
 
@@ -247,7 +247,7 @@ def get_kube_config() -> str:
 
 def get_kubectl_cmd() -> str:
     """Return the kubectl binary location within the snap, used to build k8s commands."""
-    kubectl_cmd = "{}/kubectl".format(os.environ["SNAP"])
+    kubectl_cmd = "{}/opt/spark/kubectl".format(os.environ["SNAP"])
     return kubectl_cmd
 
 
--- a/spark_client/cli/service_account_registry.py
+++ b/spark_client/cli/service_account_registry.py
@@ -110,7 +110,7 @@ if __name__ == "__main__":
 
     context = args.context or kube_interface.context_name
 
-    logging.info(f"Using K8s context: {context}")
+    logging.debug(f"Using K8s context: {context}")
 
     registry = K8sServiceAccountRegistry(kube_interface.with_context(context))
 
@@ -126,7 +126,7 @@ if __name__ == "__main__":
 
     elif args.action == Actions.DELETE:
         user_id = build_service_account_from_args(args).id
-        logging.info(user_id)
+        logging.critical(user_id)
         registry.delete(user_id)
 
     elif args.action == Actions.ADD_CONFIG:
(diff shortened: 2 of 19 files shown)


spark_client/cli/service_account_registry.py
```diff
@@ -126,7 +126,7 @@ def create_service_account_registry_parser(parser: ArgumentParser):
 
     elif args.action == Actions.DELETE:
         user_id = build_service_account_from_args(args).id
-        logging.info(user_id)
+        logging.critical(user_id)
```
deusebio, line 129: **question** why critical? Info seems fine. This does not seem something very critical to log out

spark_client/cli/service_account_registry.py
```diff
@@ -170,7 +170,7 @@ def create_service_account_registry_parser(parser: ArgumentParser):
         if maybe_service_account is None:
             raise NoAccountFound(input_service_account.id)
 
-        maybe_service_account.configurations.log(logging.info)
+        maybe_service_account.configurations.log(logging.critical)
```
deusebio, line 173: as before

helpers/utils.py
```diff
@@ -247,7 +247,7 @@ def get_kube_config() -> str:
 
 def get_kubectl_cmd() -> str:
     """Return the kubectl binary location within the snap, used to build k8s commands."""
-    kubectl_cmd = "{}/kubectl".format(os.environ["SNAP"])
+    kubectl_cmd = "{}/opt/spark/kubectl".format(os.environ["SNAP"])
```
deusebio, line 250: I'm just wondering whether kubectl should really be under opt/spark or rather some other places? I mean...it is really not connected to spark, is it? @taurus-forever ? Any suggestion?

spark_client/cli/service_account_registry.py
```diff
@@ -183,11 +183,11 @@ def create_service_account_registry_parser(parser: ArgumentParser):
         if maybe_service_account is None:
             raise NoAccountFound()
 
-        # maybe_service_account.configurations.log(logging.info)
-        logging.info(maybe_service_account.id)
+        # maybe_service_account.configurations.log(logging.critical)
+        logging.critical(maybe_service_account.id)
```
deusebio, line 187: As before, and all the rest of the logging is worth reverting

tests/unittest/test_helpers.py
```diff
@@ -232,11 +232,11 @@ def side_effect(*args, **kwargs):
 
         label = helpers.utils.get_primary_label()
 
-        cmd_get_service_account = f"{test_id}/kubectl --kubeconfig {kubeconfig} --namespace {namespace} --context {context}  get serviceaccount -l {label} -A -o yaml"
+        cmd_get_service_account = f"{test_id}/opt/spark/kubectl --kubeconfig {kubeconfig} --namespace {namespace} --context {context}  get serviceaccount -l {label} -A -o yaml"
```
deusebio, line 235: maybe it is worth factoring out the kubectl location into a single variables, just to have this in a single place

deusebio (commented):
Some comments and a double-check from @taurus-forever I'd like to validate

spark_client/cli/service_account_registry.py
averma-canonical: Actually we absolutely need to print this to the console but are using the logging interface for it.

Since the default log level is ERROR, this will never get printed out to console unless we use log level ERROR or CRITICAL.

Earlier we were using bare print statements instead of the logging interface.

spark_client/cli/service_account_registry.py
averma-canonical: explanation same as shared in the previous comment.

spark_client/cli/service_account_registry.py
averma-canonical: as explained in the previous comment, we might need to 
- either revert to the bare print statement or
- make default log level as INFO or DEBUG and use INFO for these statements.

tests/unittest/test_helpers.py
averma-canonical: :+1: 

Will push the changes once kubectl location is confirmed.

spark_client/cli/service_account_registry.py
deusebio: Discussed with @welpaolo, we should substitute the logging in "cli" files with "print", possibly adding some useful messages

deusebio:
closing as this is superseeded by #64

PR #59, opened 2023-05-04
Fixing bugs 1906, 1907, 1908 - regressions caused when porting to lightkube implementation

averma-canonical:
After porting to lightkube some regressions were found, mainly printing of user relevant info in the cli tool as well as listing all service accounts which is a base case for many other operations.

Also, avoiding fetching latest spark version all the time. Instead always work with spark version approved as present in SPARK_VERSION file in the repo.

spark-client version 0.0.7 has already been used and was the one with botched lightkube implementation. Hence using tag 0.0.8 for the spark-client version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ follow_imports = "silent"
 
 [tool.poetry]
 name = "spark-client"
-version = "0.0.6"
+version = "0.0.8"
 description = "This project allow you to create a SNAP for interacting with a K8s cluster to submit Spark jobs or run spark shells interactively"
 authors = [
     "Abhishek Verma <abhishek.verma@canonical.com>",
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -34,7 +34,7 @@ follow_imports = "silent"
 
 [tool.poetry]
 name = "spark-client"
-version = "0.0.6"
+version = "0.0.8"
```
deusebio, line 37: we have a bit of a mess with versioning of packages. We used to have a release branch that was tagged in order to release a new python package. It is fine to upgrade this version here, but please do not tag this commit, but let's use the release branch for this.

deusebio (approved):
Changes are sane. 

We will restructure this repo a bit in the next couple of weeks but it is good to start that from a stable point (which is this PR)

welpaolo (approved):

PR #60, opened 2023-05-05
Fix Spark version mismatch

welpaolo:
This PR addresses the inconsistencies between the rock image and stabilize the tests.


tests/integration/ie-tests.sh
```diff
@@ -203,18 +203,21 @@ teardown_test_pod() {
...
-  SPARK_EXAMPLES_JAR_NAME='spark-examples_2.12-3.3.2.jar'
+  SPARK_EXAMPLES_JAR_NAME='spark-examples_2.12-3.4.0.jar'
 
   PREVIOUS_JOB=$(kubectl get pods | grep driver | tail -n 1 | cut -d' ' -f1)
 
   NAMESPACE=$1
   USERNAME=$2
 
-  kubectl exec testpod -- env UU="$USERNAME" NN="$NAMESPACE" JJ="$SPARK_EXAMPLES_JAR_NAME" \
+  SPARK_IMAGE='ghcr.io/canonical/charmed-spark:3.4.0-22.04_edge'
```
deusebio, line 213: **niptick** Why not factoring this out in a single place at the top of the file? I can see people changing this only in one place and not another...

deusebio (approved):
LGTM!

tests/integration/ie-tests.sh
welpaolo: Done

tests/integration/ie-tests.sh
```diff
@@ -18,7 +20,7 @@ run_example_job() {
   KUBE_CONFIG=/home/${USER}/.kube/config
 
   K8S_MASTER_URL=k8s://$(kubectl --kubeconfig=${KUBE_CONFIG} config view -o jsonpath="{.clusters[0]['cluster.server']}")
-  SPARK_EXAMPLES_JAR_NAME='spark-examples_2.12-3.3.2.jar'
+  SPARK_EXAMPLES_JAR_NAME='spark-examples_2.12-3.4.0.jar'
```
juditnovak, line 23: Would it be possible somehow to reduce the number of all these duplications...?
Across as many places as possible (config, yaml, etc...)
I'm wondering about a common source like the `SPARK_VERSION` file, or at least as few re-usable sources as we could have...?

Of course if we're already at the strict minimum (due to various technologies being unable to share info across themselves) then ignore my comment...

juditnovak: (Sry, unclear without hte context of the rest of the script. What I was aiming to refer to was the same version number (change) that had to be performed many times across the same or different scripts, config files, github workflow yaml-s.... Kinda leaving much opportunity for an 'inconsistency' type of mistake.

juditnovak: For the record: we've discussed the above question with @welpaolo , and agreed that that the change is necessary but to be done outside of this "urgent fix" PR scope. We opened Jira ticket (DPE-1934)[https://warthogs.atlassian.net/browse/DPE-1934] to track the issue

juditnovak (approved):
LGTM

PR #62, opened 2023-05-19
[DPE-1948] new lib fetched by snap

juditnovak:
Enabling the snap build to rely on the `spark8t` toolkit (code that was local to the snap before).

Special thanks to @welpaolo for his patient and extremely useful help.


snap/snapcraft.yaml
```diff
@@ -58,16 +61,23 @@ apps:
 
 parts:
   spark:
-    plugin: nil
+    plugin: python
+    python-packages:
+      # - https://github.com/canonical/spark-k8s-toolkit-py/releases/download/v0.0.1/spark8t-${SPARK8T_VERSION}-py3-none-any.whl
```
welpaolo, line 66: Please remove the commented line :)

snap/snapcraft.yaml
```diff
@@ -16,29 +16,32 @@ plugs:
 
 environment:
   JAVA_HOME: $SNAP/usr/lib/jvm/java-11-openjdk-amd64
-  PATH: $JAVA_HOME/bin:$PATH
+  PATH: $JAVA_HOME/bin:$PATH:$SNAP
+  KUBECONFIG: $SNAP_REAL_HOME/.kube/config
+  SPARK_HOME: $SNAP
+  SPARK_CONFS: $SNAP/lib/python3.10/site-packages/spark8t/resources/conf/
+  SPARK_USER_DATA: $HOME
+  SPARK8T_VERSION: 0.0.1
```
welpaolo, line 24: Why do we need this variable ?

poetry.lock
welpaolo, line 1: I think we want to keep both `poetry.lock` file and `requirements.txt` ? @deusebio can you confirm ?

snap/snapcraft.yaml
```diff
@@ -112,39 +122,21 @@ parts:
         cp -r jars/* $CRAFT_PART_INSTALL/jars/
         mkdir -p $CRAFT_PART_INSTALL/python
         cp -r python/* $CRAFT_PART_INSTALL/python/
+
+        echo "Installing spark8t"
```
welpaolo, line 126: I think this line could be removed

welpaolo (approved):
Great work Judit! I left some minor comments

snap/snapcraft.yaml
deusebio: **suggestion** I don't mind too much. I mean, once we are confident that the refactoring is stable, we should probably revert on using the tar.gz instead of the branch. Up to @juditnovak (being a suggestion)

snap/snapcraft.yaml
deusebio, line 126: **todo** I believe it is best (and recommended) to keep the different components separated in different `parts` ([ref](https://snapcraft.io/docs/adding-parts)). I'd install the `spark8t` package in a separate `part` as it was originally done before. I see, as the two (spark binaries and spark8t) are indeed two different components.

snap/snapcraft.yaml
```diff
@@ -112,39 +122,21 @@ parts:
...
+        chmod -R 755 $CRAFT_PART_INSTALL/lib/python3.10/site-packages/spark8t/cli/
     override-prime: |
         snapcraftctl prime
         rm -vf usr/lib/jvm/java-11-openjdk-*/lib/security/blacklisted.certs
 
   helper-package:
```
deusebio, line 132: **suggestion** Now that we understand snap a bit better, maybe it makes sense to divide the snap into 3 parts:
* spark
* spark8t
* kubectl

I feel these are indeed the components of our snap. @welpaolo: how about that?

tests/examples.py
```diff
@@ -1,7 +1,7 @@
 import os
```
deusebio, line 1: **todo** Remove file

tox.ini
```diff
@@ -7,16 +7,14 @@ skip_missing_interpreters = True
 envlist = lint, unit
```
deusebio, line 7: **todo** I don't think we need these anymore, given we really should not have any python code anymore

deusebio (changes requested):
Great start! 

I have few comments/suggestions, but mostly I believe that we should probably clean all the code/files linked to the python bits (since we have now externalized all python code in a separated repository), like:
* `poetry.lock` and `requirements.txt`
* `pyproject.toml`
* `tox.ini` (keeping the makefile for build, installing, integration testsing - which is not python related)
* helpers (I noticed that we still have the helpers folder which we should definitely get rid of)
* python files in the `tests` folder (except the `test-pyspark.py`, which is needed for the integrations tests)

I also propose a (hopefully) small refactoring ([this](https://github.com/canonical/spark-client-snap/pull/62#discussion_r1204772366) comment) to use snapcraft [parts](https://snapcraft.io/docs/adding-parts) to highlight in the structure the different component that we have

poetry.lock
juditnovak: This was by mistake, adding it back.

snap/snapcraft.yaml
juditnovak: So THIS is why we would need that variable.

Frankly, I don't mind at all what we wanna do on this. I just thought that soon enough we'll switch to a versioned, packaged reference to the `spark8t` package instead of the current, pretty fragile solution taking the latest `main/HEAD`.

The line above, together with the variable pointing to the desired `spark8t` version is to facilitate this future step.

I can either add a comment above, like 
```
# Location of the spark8t package for future reference
```
or I can completely remove it, I really don't mind/care. This was just to facilitate a soon upcoming change.

snap/snapcraft.yaml
juditnovak: See later in https://github.com/canonical/spark-client-snap/pull/62/files/8e036a644350a62f175248f181cfd36e24104a7c#r1205100357 :-)

I'm good to actually comment out the variable, with an added comment above (as suggested in the related comment). Or to remove it completely if you Guys don't want this.

We will want to pin down the package version though, I don't think that's avoidable.... And we may not only want to get the latest version of `spark8t` but rahter a 'stable' one (avoiding a chain of broken snap/roc unexpectidly). So I think we'll need some level of "manual" control on this, instead of a fully automated chain....?

Anyway, again, I do whatever you Guys want to have.

snap/snapcraft.yaml
juditnovak: Bah, indeed, good point.

poetry.lock
deusebio: No, wait. We don't want to keep these. I don't see any reason to keep these files, since they are now stored elsewhere

snap/snapcraft.yaml
juditnovak: If you preferer so.
Spark binaries are strongly bound to `spark8t`, so I don't think this would be a mess (one is incomplete w/o the other one in our context.) But if preferred I can split the 2 steps.

snap/snapcraft.yaml
deusebio: yes, the point here is that this is a variable that you need at build time (since it refers to the tarball do be downloaded) but not at execution, therefore it is not needed among those env variables.

snap/snapcraft.yaml
deusebio: Let's keep it such that the comments will hurt our OCD around these things, and we will switched to the versioned one sooner rather than later

snap/snapcraft.yaml
welpaolo: I think this is a good idea, it will make the structure of the snap much cleaner.

snap/snapcraft.yaml
```diff
@@ -16,29 +16,32 @@ plugs:
 
 environment:
   JAVA_HOME: $SNAP/usr/lib/jvm/java-11-openjdk-amd64
-  PATH: $JAVA_HOME/bin:$PATH
+  PATH: $JAVA_HOME/bin:$PATH:$SNAP
+  KUBECONFIG: $SNAP_REAL_HOME/.kube/config
+  SPARK_HOME: $SNAP
+  SPARK_CONFS: $SNAP/lib/python3.10/site-packages/spark8t/resources/conf/
```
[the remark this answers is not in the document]
juditnovak: OK, indeed this should be re-visited. However especially since it requires changes on 2 repos, I agree to leave it for a little later, and now just let's have a "green chain" that's primarily correct, even if there's further cosmetics 2 B done.

snap/snapcraft.yaml
juditnovak: I see your point... However I'm not sure if we can define it within the scope of the corresponding build step (using the `python` plugin)... I'm afraid it has to be declared before we refer to the pkg, I can't just have it defined on the same level, no?

juditnovak: I've just checked this doesn't work:
```
parts:
  spark8t:
    python-packages:       
      - https://github.com/canonical/spark-k8s-toolkit-py/releases/download/v0.0.1/spark8t-${SPARK8T_VERSION}-py3-none-any.whl
    build-environment:                                                                                                                                                                                          
      - SPARK8T_VERSION: 0.0.1
```
What else could be the correct location of this variable then, if we don't want to have an ENV_VAR? (I kinda see the reason, it's not really withing the scope of ENV_VARs, I just don't see a conceptually better solution -- may be my ignorance on snaps?)

juditnovak: Bah, it's even worse: perhaps it's not the location of the variable but it may not be possible to add variables to the `python` plugin `python-packages` section.

Fine, I'll just remove the whole thing for now and we'll experiment when we put this in 'production'

deusebio: Yes, for the time being we can either hard-code the package name  (which is not too far from the fact that as-is we are hard coding an env-variable for the build anyway) or just drop this and address this later.

snap/snapcraft.yaml
juditnovak: See https://github.com/canonical/spark-client-snap/pull/62#discussion_r1205277780 removed for now.

snap/snapcraft.yaml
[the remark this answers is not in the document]
juditnovak: I've added a reference of this thread to https://warthogs.atlassian.net/browse/DPE-1884

snap/snapcraft.yaml
juditnovak: Done

snap/snapcraft.yaml
juditnovak: Done

tests/examples.py
juditnovak: done

Makefile
deusebio, line 1: **niptick** There are still a couple of references here to be cleaned in this file

* Line 3
* Line 9 and 10 (these variables are not used anymore)
* Line 20 (POETRY_EXISTS is not relevant anymore)
* Line 23 (probably can be taken away, but just worth trying out and seeing whether CI still passes)

snap/snapcraft.yaml
```diff
@@ -112,39 +126,19 @@ parts:
         cp -r jars/* $CRAFT_PART_INSTALL/jars/
         mkdir -p $CRAFT_PART_INSTALL/python
         cp -r python/* $CRAFT_PART_INSTALL/python/
+
     override-prime: |
         snapcraftctl prime
         rm -vf usr/lib/jvm/java-11-openjdk-*/lib/security/blacklisted.certs
 
   helper-package:
-    plugin: python
```
deusebio, line 120: **suggestion** What about changing the name from `helper-package` to `kubectl`?

deusebio (approved):
Great work! I really like how the snap repo has now shaped up. 

I just have a couple of small minor suggestions/nipticks but they don't warrant approval. Once you fix them, feel free to merge straight away

Makefile
juditnovak: Good call, indeed I've overlooked those references. Furthermore, while removing those, I found a few more small, harmless things :sweat_smile: 

Done now (tests OK locally, waiting 4 the Neverending Pipeline)

snap/snapcraft.yaml
juditnovak: Makes sense, done :-)

PR #67, opened 2023-06-12
import from discourse

deusebio:
This PR results from syncing of discourse after @juditnovak review and updates, i.e. ticket [DPE-2022](https://warthogs.atlassian.net/browse/DPE-2022)

[DPE-2022]: https://warthogs.atlassian.net/browse/DPE-2022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/how-to/h-run-on-k8s-pod.md
```diff
@@ -17,7 +17,7 @@ metadata:
 spec:
   containers:
   - name: spark-client
-    image: ghcr.io/canonical/charmed-spark:3.4.0-22.04_edge
+    image: ghcr.io/canonical/charmed-spark:3.3.2-22.04_edge
```
deusebio: This does not look correct

docs/index.md
```diff
@@ -43,6 +42,7 @@ Spark Client Snap is an official distribution of Apache Spark. It’s an open-so
 | 2     | h-use-spark-client-from-python | [Use the Spark Client Python API](/t/spark-client-snap-how-to-python-api/8958)                                       |
 | 2     | h-run-on-charmed-k8s           | [Run on Charmed Kubernetes](/t/spark-client-snap-how-to-run-on-charmed-kubernetes/8960)                              |
 | 2     | h-run-on-k8s-pod               | [Run on K8s pods](/t/spark-client-snap-how-to-run-on-k8s-in-a-pod/8961)                                              |
+| 2     | h-spark-streaming               | [Run Spark Streaming Jobs](/t/charmed-spark-how-to-run-a-spark-streaming-job/10880)                                              |
```
deusebio, line 49: We probably forgot to add this tracking line before. I have done this now

docs/tutorial/t-setup-environment.md
```diff
@@ -59,4 +59,9 @@ sudo snap install spark-client --edge
...
 
 This will install all the spark binaries along side with the spark-client utilities provided in 
-the snap.
\ No newline at end of file
+the snap.
+
+***
+
+ * Previous: [Spark Client Snap Tutorial](https://discourse.charmhub.io/t/spark-client-snap-tutorial/8957)
+ * Next: [Manage Spark service accounts](https://discourse.charmhub.io/t/spark-client-snap-tutorial-setup-environment/8952)
```
deusebio, line 64: **praise** I like this "Previous" and "Next"

docs/how-to/h-manage-service-accounts.md
```diff
@@ -1,72 +1,89 @@
...
+```
 
-### List all service accounts
+The command below creates a service account associated with configuration properties provided either via property file or explicit 
+configuration arguments.  The flags `--primary` specifies that the newly created account will be the primary account to 
+be used. (If another primary exists, the latter account primary flag will be set to `false`.)
 
 ```bash
-spark-client.service-account-registry list
+spark-client.service-account-registry create --username demouser --namespace demonamespace  --primary --properties-file /home/demouser/conf/spark-overrides.conf  --conf spark.app.name=demo-spark-app-overrides
```
grobbie, line 25: if the account already exists on the cluster, what will happen?

docs/how-to/h-manage-service-accounts.md
```diff
@@ -1,72 +1,89 @@
...
 ```bash
 spark-client.service-account-registry get-primary
 ```
 
-This command will allow the user to find out which one is the primary account, providing information 
-about its configuration settings. 
-
 ### Cleanup a Service Account
 
+This command will delete the service account together with the other resources created, e.g.
```
grobbie, line 81: I think we should explain that this tool can be used by the Kubernetes cluster admin in order to create the correct resources for the spark application to run. That can be separate from the user who actually uses spark on the cluster in an interactive way. We need to explain this separation of duties.

juditnovak (approved):
May not be perfect but better than what it was :-)

docs/how-to/h-run-on-k8s-pod.md
```diff
@@ -1,73 +1,72 @@
...
-after that, one logs in with
+We can log into the pod as below
 
 ```shell
-$ kubectl exec --stdin --tty shell-demo -n spark-test-ns -- /bin/bash 
+$ kubectl exec --stdin --tty shell-demo -- /bin/bash 
 ```
 
-Now from within the pod, lets create the kubeconfig for use with spark-client. ```KUBECONFIG_CONTENTS_FROM_CHARMED_KUBERNETES``` will come from the first step executed outside the Charmed Kubernetes cluster pod. 
+Now let's create the Kubernetes configuration on the pod, with contents from the original server's `.kube/config`
```
grobbie, line 38: is this step necessary? I think all the kubeconfig info is made available to the pod in downwards API

docs/how-to/h-use-spark-client-from-python.md
```diff
@@ -121,8 +127,8 @@ registry.delete(service_account.id)
 
 #### Manage Primary Accounts
 
-In the Spark Client Snap, there is the notion of the primary service account, the 
-one that would be chosed by default, if no specific information is provided. The
+Spark has a notation of the so called 'primary' service account, the
```
grobbie: spark8t has this notion, spark doesn't afaik

docs/tutorial/t-setup-environment.md
```diff
@@ -1,20 +1,17 @@
...
-To install MicroK8s
+MicroK8s snap installation goes as simple as
 
 ```bash
 sudo snap install microk8s --classic
 ```
 
-It is also recommended to configure MicroK8s, so that the user has admin permission on 
-the cluster. Use the following commands to configure the user to perform actions on the K8s cluster:
+It is also recommended to configure MicroK8s in a way, so that the correpsonding user has admin rights on  the cluster.
```
grobbie: spelling - "corresponding"

grobbie: also, this needs further explanation - why does the corresponding user need admin rights?

docs/tutorial/t-spark-service-accounts.md
```diff
@@ -73,25 +70,27 @@ options:
...
-properties file while overriding the configuration properties ```spark.app.name``` and ```spark.executor.instances```.
-
-For [job submission](https://discourse.charmhub.io/t/spark-client-snap-tutorial-spark-submit/8953), this service account along with it's default configuration properties can be used to submit a Spark job. 
+The above command sets up a service account for user ```demonamespace:demouser``` for Spark job submission. Configuration properties are taken from the specified 
+properties file, potentially overriden by configuration properties ```spark.app.name``` and ```spark.executor.instances```.
 
-For example, assuming the properties file provided has configuration details to access data in S3, one could submit a job like
-```bash
-spark-client.submit  --username demouser --namespace demonamespace --deploy-mode cluster --conf spark.app.name=demo-spark-s3-app $S3_PATH_FOR_CODE_FILE
+**_Note:_** The command described above does not create a kubernetes namespace but needs it to be there. It does however create the requested username in the specified and existing namespace. The kubernetes namespace could be created such as:
```
grobbie, line 85: perhaps we should expand the functionality of the tool over time to allow for namespace creation?

grobbie (approved):
very good - left a few comments and suggestions that you might like to fix

docs/how-to/h-manage-service-accounts.md
deusebio: This command would fail actually, with an exception. However, the output is not too clear and it is indeed a rather common use-caes and exception that should be handled appropriately. I have raised an issue to be addressed in the next pulses. When addressing this, we will also update the documentation [here](https://github.com/canonical/spark-k8s-toolkit-py/issues/23)

docs/how-to/h-run-on-k8s-pod.md
deusebio: Yes, I believe so. We have a ticket in the backlog, [DPE-1491](https://warthogs.atlassian.net/browse/DPE-1491), but this has not been addressed yet

docs/how-to/h-use-spark-client-from-python.md
deusebio: Revised the text!

docs/tutorial/t-spark-service-accounts.md
deusebio: Yes, indeed. I have created an associated ticket ([DPE-2141](https://warthogs.atlassian.net/browse/DPE-2141)) about this

docs/how-to/h-manage-service-accounts.md
deusebio: I have addressed this in a number of places in the text

grobbie (approved):
thanks

Issue #68, opened 2023-06-14
spark-shell executor pods won't start up with latest (post 2023.06.08)


mentics:
## Steps to reproduce
1. Try to run spark-client.spark-shell

## Expected behavior
Executor pods start up.

## Actual behavior

They fail repeatedly, with error in the logs:
`error: unknown command "executor", see 'pebble help'`

## Versions

Operating system: Ubuntu 22.04.2 LTS

## Log output

```error: unknown command "executor", see 'pebble help'```

## Additional context

It was working last week and then I added a new node to my cluster that pulled the new image. Then the nodes using the old image worked, but those using the new image did not. It seems likely the problem is in the image: ghcr.io/canonical/charmed-spark:3.4.0-22.04_edge

deusebio:
Hi @mentics ! Thanks for reporting this. Yes, we have recently (today) upgraded the [spark image](https://github.com/canonical/charmed-spark-rock/pull/23) to use pebble as entrypoint, which is now the new standard entrypoint required by ROCK-compliant OCI images. 

This change unfortunately is not backward compatible (as changes the entrypoint) and broke things on the spark client snap side (we are still on edge risks here). I'm going to fix this shortly. I'll refer this issue in the PR so you can track when this is resolved

deusebio:
As a temporarily workaround you can also:

1. Create a podTemplate.yaml like [this](https://github.com/canonical/charmed-spark-rock/pull/23/files#diff-de9a13b19681e3a36d605f5fd2939855883d693d61ad2e5433bd6f240c9bc93c) one. 
2. Refer to that file in the configuration, similarly to what has been done [here](https://github.com/canonical/charmed-spark-rock/pull/23/files#diff-42c4bbc2d848486731ee3dacc12ba05a9be8df2936f2cfaea32f643e73994bc6). You can either modify the `/var/snap/spark-client/41/etc/spark8t/spark-defaults.conf` file or you can provide those settings via the --conf arguments in the cli

mentics:
> As a temporarily workaround you can also:
> 
> 1. Create a podTemplate.yaml like [this](https://github.com/canonical/charmed-spark-rock/pull/23/files#diff-de9a13b19681e3a36d605f5fd2939855883d693d61ad2e5433bd6f240c9bc93c) one.
[...]

PR #69, opened 2023-06-14
[DPE-2108] Fix issue #68

deusebio:
Addressing issue #68 related to new upgrade of Spark ROCK image, using pebble entrypoint


grobbie (approved):
Looks ok technically, but as discussed offline I'm concerned about the UX we have introduced with pod defaults. Approving for now, but we may revisit this.

snap/local/spark8t/spark-defaults.conf
```diff
@@ -0,0 +1,3 @@
+spark.kubernetes.container.image=ghcr.io/canonical/charmed-spark:3.4.0-22.04_edge
```
juditnovak: Hmmm, I'm confused... This was supposed to be here already...

snap/snapcraft.yaml
```diff
@@ -63,6 +63,12 @@ parts:
   spark8t-conf:
     plugin: dump
     source: snap/local
+    organize:
+      spark8t/spark-defaults.conf: etc/spark8t/spark-defaults.conf
+      spark/podTemplate.yaml: opt/spark/conf/podTemplate.yaml
+    stage:
+      - etc/spark8t/spark-defaults.conf
+      - opt/spark/conf/podTemplate.yaml
```
juditnovak: Instead of "hacking" `podTemplate.yaml` into another place than the original `local` structure -- don't we wanna have a clear sturcture on `local`, precisely reflecting the delivery (I mean packed snap)? I think it would be both more intuitive and clean.

Practically what I mean is to supress all the new code above, and 
```
mkdir snap/local/spark/conf/
git mv snap/local/spark/podTemplate.yaml snap/local/spark/conf/podTemplate.yaml
```

snap/local/spark8t/spark-defaults.conf
deusebio: yes, but git works with blobs which is not exactly line by line. Git is probably seeing two different blobs, therefore does not mention the fact that the file was moved.

juditnovak: All right, so the first line has already been here, we only add line 2+3 to this file?
The rest is original content in the original place?

snap/snapcraft.yaml
deusebio: I don't think the path on "local" matters too much, given the organize flag. I kinda avoided creating very nested directories (it would be /snap/local/opt/spark/conf, but yes, it is clearer to map to the same structure. And we don't pay per number of directories, do we?

juditnovak: Yeah, generally I agree to keep structures as flat as possible.
But here I rather prefer to keep consistency between "code" (repo) structure, and the actual deployment -- and use the `dump` plugin with "no additional magic" but only copying a clear, consistent structure over.

juditnovak (approved):

deusebio:
superseded by #73 (given we have been rolling back to the previous Pebble/Rockcraft UX on the OCI image)

PR #70, opened 2023-06-21
[DPE-1909] History Server charm deployment how-to

deusebio:


docs/how-to/h-deploy-spark-history-server.md
```diff
@@ -0,0 +1,96 @@
+## Deploy the Spark History Server 
+
+The Spark History server is a component in the Spark ecosystem that helps you monitor your spark jobs. In particualar, the History Server enables access to logs of completed/running Spark jobs for viewing and analysis.
```
grobbie: capitalise Spark in "Spark jobs"

docs/how-to/h-deploy-spark-history-server.md
```diff
@@ -0,0 +1,96 @@
...
+
+Within the Charmed Spark solution, its deployment, configuration and operation is handled by [Juju](https://juju.is/). Therefore, a [Charmed operator](https://charmhub.io/spark-history-server-k8s) will be responsible for its management and operations on K8s.
```
grobbie: Maybe say "Therefore, a Juju charm(link to charm) will be responsible for its management and operations on K8s." Then provide a short explanation of what a Juju charm is. I think there is a standard explanation text that Massi and Jon recently prepared.

grobbie: Maybe also worth mentioning that there's a Juju provider for Terraform that users can leverage to integrate Juju with their existing systems management and provide a link. We should probably do that in all of our documentation.

docs/how-to/h-deploy-spark-history-server.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Requirements
+
+Since Spark History server will be managed by Juju, make sure that:
+* you have a Juju client (e.g. via a [SNAP](https://snapcraft.io/juju)) installed in your local machine  
+* you are able to connect to a juju controller running on K8s.
+* you have read-write permissions (therefore you have a access key and access secret) to a S3-compatible object storage (e.g. MinIO)
+
+To see how to setup a Juju controller on K8s and the juju client, you can refer to existing tutorials and documentation, e.g. [here](https://juju.is/docs/olm/get-started-with-juju).
```
grobbie, line 14: add "and to learn more about Juju" or similar words

docs/how-to/h-deploy-spark-history-server.md
```diff
@@ -0,0 +1,96 @@
...
+
+To see how to setup MinIO on top of MicroK8s, please refer to [here](https://microk8s.io/docs/addon-minio). If you have a MicroK8s MinIO plugin enabled, use the following commands to 
+obtain the access key, the access secret and the MinIO endpoint:
+
+* *access key*: `microk8s.kubectl get secret -n minio-operator microk8s-user-1 -o jsonpath='{.data.CONSOLE_ACCESS_KEY}' | base64 -d`
+* *secret key*: `microk8s.kubectl get secret -n minio-operator microk8s-user-1 -o jsonpath='{.data.CONSOLE_SECRET_KEY}' | base64 -d`
+* *MinIO endpoint*: `microk8s.kubectl get services -n minio-operator | grep minio | awk '{ print $3 }'`
```
grobbie, line 21: You could mention here that users don't have to use MinIO or MicroK8s - they can also use S3 and other Kubernetes, but the commands may differ in that case.

docs/how-to/h-deploy-spark-history-server.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Preparation
+
+#### Juju Namespace 
+
+Before deploying the Spark History server, it is advisable to create a dedicated namespace for `Spark` components if you don't already have one:
```
grobbie: It's a requirement of Juju to have a namespace to deploy it's application. Need to be more specific - it's not optional.

docs/how-to/h-deploy-spark-history-server.md
```diff
@@ -0,0 +1,96 @@
...
+
+```bash 
+juju add-model spark
+```
```
grobbie, line 34: add-model might be counterintuitive and need a little explanation that this command creates a namespace in K8s that Juju can use.

docs/how-to/h-deploy-spark-history-server.md
```diff
@@ -0,0 +1,96 @@
...
+s3.create_bucket(Bucket="history-server")
+s3.put_object(Bucket="history-server", Key=("spark-events/"))
+```
+
+### Deploy Spark History Server
+
+* *Deploy the Spark History Server component*
+
+```bash
+juju deploy spark-history-server-k8s -n1 --channel 3.4/edge
```
grobbie, line 68: is risk mandatory? could they deploy `--channel 3.4` without specifying the risk?

docs/how-to/h-deploy-spark-history-server.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+* *Deploy and configure a [s3-integrator](https://charmhub.io/s3-integrator) charm*
+
+```bash
+juju deploy s3-integrator -n1 --channel edge
+juju config s3-integrator bucket="history-server" path="spark-events" endpoint=<S3_ENDPOINT>
+juju run s3-integrator/0 sync-s3-credentials access-key=<ACCESS_KEY> secret-key=<SECRET_KEY>
```
grobbie: these commands need more explanation. -n1 means deploy one pod of s3-integrator. `juju run` runs one of a set menu of commands on that pod.

grobbie: is it `run` or `run-action`?

docs/how-to/h-deploy-spark-history-server.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+* *Relate s3-integrator and Spark History server*
```
grobbie, line 79: needs more explanation. What is happening here?

docs/how-to/h-deploy-spark-history-server.md
```diff
@@ -0,0 +1,96 @@
...
+
+```bash 
+juju relate s3-integrator spark-history-server-k8s
+```
+
+Make sure that the s3-integrator and Spark History server services are active using `juju status`.
```
grobbie: provide a screenshot?

grobbie (changes requested):
a few bits and bobs to change here and there

docs/how-to/h-deploy-spark-history-server.md
deusebio: Uhm, not sure about this. I mean, this is a how-to guide where we should not provide either explanations or other extra information than a step by step guide to get their task done. Actually I definitely agree that it would be good to provide a terraform template, but probably in a separate topic that this one

docs/how-to/h-deploy-spark-history-server.md
deusebio: Uhm, not sure. I would not talking about `learning` that would rather suit a "tutorial" rather than a a "how-to"

docs/how-to/h-deploy-spark-history-server.md
deusebio: My point was that it is advisable (but optional) to have a namespace/model for Spark components ONLY. Not that it is advisable to have a model in general. I'll try to clarify this.

docs/how-to/h-deploy-spark-history-server.md
deusebio: I believe for `edge` it needs to be specified. If not specified, it will probably fall back to stable (when we have one). Note that this is also the suggestion to deploy the charm provided in [charmhub](https://charmhub.io/spark-history-server-k8s?channel=3.4/edge)

docs/how-to/h-deploy-spark-history-server.md
deusebio: Uhm, I don't think we need to provide explanation in this how-to, but just provide a bunch of commands to copy and paste, according to the diataxis framework

deusebio: Uhm, I don't think we need to provide explanation in this how-to, but just provide a bunch of commands to copy and paste, according to the diataxis framework

deusebio: > is it run or run-action?

Great catch. I usually switch to the new juju ux, the v2 (that I believe should become the new standard in juju3). But it is best to keep things consistent with juju 2.9.x that is the one we are still on with this charm

docs/how-to/h-deploy-spark-history-server.md
deusebio: Also here, the users in how-to should not be interested what is happening under the hood. I have nevertheless added a line that after this steps the spark history server should be able to read from s3.

deusebio:
I have addressed the comments (where relevant) and also added another extra section that I felt it was missing about configuring the Spark service account to write logs to s3

PR #77, opened 2023-08-18
[MISC] Change default branch on scheduled workflows

deusebio:
The scheduled process was still poitint at main branch, which had been deleted, hence the CI was failing. 

Meanwhile, I have also improved/cleaned a bit the script making it parametric with respect of base and update branches


welpaolo (approved):
LGTM

marcoppenheimer (approved):

PR #84, opened 2023-10-26
[DPE-2784] Update Spark-k8s-toolkit library

welpaolo:


snap/snapcraft.yaml
```diff
@@ -67,7 +67,7 @@ parts:
   spark8t:
     plugin: python
     python-packages:
-        - https://github.com/canonical/spark-k8s-toolkit-py/releases/download/v0.0.1/spark8t-0.0.1-py3-none-any.whl
+        - https://github.com/canonical/spark-k8s-toolkit-py/releases/download/v0.0.2/spark8t-0.0.2-py3-none-any.whl
```
deusebio, line 70: **niptick** Maybe we could factor out the spark8t version into one variable to be exposed as env var. This is clearly not blocking. If you have 10 min to try this out fine, otherwise we can approach this on a dedicated PR, as it goes a bit beyond the scope

deusebio (approved):
LGTM

PR #91, opened 2024-02-15
[DPE-2164] Expose Spark SQL in Spark Client snap

theoctober19th:
This PR exposes the `spark-sql` binary in the `spark-client` snap and adds integration tests for it.


.github/workflows/ci.yaml
```diff
@@ -49,6 +49,11 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v4
 
+      - name: Setup MicroK8s
+        run: |
+          make microk8s
```
theoctober19th, line 54: I had to add this step because `sg microk8s` in the next step would not work.

tests/integration/config-aws-cli.sh
```diff
@@ -7,11 +7,29 @@ sudo snap install aws-cli --classic
...
+
 DEFAULT_REGION="us-east-2"
 
 # Configure AWS CLI credentials
 aws configure set aws_access_key_id $ACCESS_KEY
 aws configure set aws_secret_access_key $SECRET_KEY
 aws configure set default.region $DEFAULT_REGION
+aws configure set endpoint_url "http://$S3_ENDPOINT"
+
+until aws s3 ls &> /dev/null; do
```
theoctober19th: I noticed that it takes a brief amount of time (something like 30 seconds or so) for the`aws cli` to work after setting up the credentials. I am not sure of what is causing that thing to happen at this moment (because this wasn't the case before). I spent a couple of days trying to understand this, and later opted for this workaround of waiting until, `aws s3 ls` command works.

welpaolo (approved):
LGTM

.github/workflows/ci.yaml
welpaolo: Weird...

tests/integration/config-aws-cli.sh
welpaolo: you can also put some `sleep` to give the time to microk8s to properly start. 
I guess that wait was hide in the the `get_endpoint` function

theoctober19th:
I have added a retry logic just in case that the runner is not left waiting forever.

tests/integration/config-aws-cli.sh
```diff
@@ -7,11 +7,49 @@ sudo snap install aws-cli --classic
...
 DEFAULT_REGION="us-east-2"
 
 # Configure AWS CLI credentials
 aws configure set aws_access_key_id $ACCESS_KEY
 aws configure set aws_secret_access_key $SECRET_KEY
 aws configure set default.region $DEFAULT_REGION
+aws configure set endpoint_url "http://$S3_ENDPOINT"
+
+
+retries=0
```
deusebio: **niptick** you could may be just put this in a function, just for the sake of having a clean, minimal main

deusebio (approved):
LGTM!

welpaolo (approved):

tests/integration/config-aws-cli.sh
theoctober19th: Done.

tests/integration/config-aws-cli.sh
theoctober19th: Added a definite retry logic.

PR #100, opened 2024-06-13
[DPE-4482] Add tests for Azure in Spark Client Snap

theoctober19th:


tests/integration/ie-tests.sh
```diff
@@ -181,17 +146,12 @@ run_pyspark_s3() {
...
   # First create S3 bucket named 'test'
-  create_s3_bucket test
+  create_s3_bucket $S3_BUCKET
 
   # Copy 'example.txt' script to 'test' bucket
-  copy_file_to_s3_bucket test ./tests/integration/resources/example.txt
-
-  echo -e "$(cat ./tests/integration/resources/test-pyspark-s3.py | spark-client.pyspark \
+  copy_file_to_s3_bucket $S3_BUCKET ./tests/integration/resources/example.txt
+  echo -e "$(sed "s/S3_BUCKET/$S3_BUCKET/g" ./tests/integration/resources/test-pyspark-s3.py | spark-client.pyspark \
```
welpaolo: A little bit tricky here :D

welpaolo (approved):
LGTM. 

Same comment as the other PR. We have too much bash and it is duplicated across multiple repos. As future work we should use more the `spark_test` lib

deusebio (commented):
Similarly to the other PR, it is a bit weird that we test different things based on the object storage. So for instance:

* S3 tests spark-sql and pyspark
* Azure tests spark-submit

Could we refactor this such that we divide setup and the actual trigger? This way we could easily compose the tests into testing spark-sql, pyspark and spark-submit on both s3 and azure (hence 6 tests). 

In general, my expectation is that we suppose azure object storage on all three commands. Would there be any issue with this?

theoctober19th:
> Similarly to the other PR, it is a bit weird that we test different things based on the object storage. So for instance:
> 
> * S3 tests spark-sql and pyspark
> * Azure tests spark-submit
> 
> Could we refactor this such that we divide setup and the actual trigger? This way we could easily compose the tests into testing spark-sql, pyspark and spark-submit on both s3 and azure (hence 6 tests).
> 
> In general, my expectation is that we suppose azure object storage on all three commands. Would there be any issue with this?

Thanks @deusebio. I have refactored the tests such that azure storage is tested on all three commands.

tests/integration/ie-tests.sh
```diff
@@ -1,6 +1,17 @@
...
 
-readonly SPARK_IMAGE='ghcr.io/canonical/charmed-spark:3.4-22.04_edge'
+# Copyright 2024 Canonical Ltd.
+
+# Import reusable utilities
+source ./tests/integration/utils/s3-utils.sh
+source ./tests/integration/utils/azure-utils.sh
+
+
+readonly SPARK_IMAGE='ghcr.io/canonical/charmed-spark@sha256:fd458fbbe4b90232ec657c92375da01789396f914865bcb4c0f919bf33c031f4'
```
deusebio: **todo** now that the debugging is done, we can revert to NOT pin the image, and use `3.4-22.04_edge`

tests/integration/ie-tests.sh
```diff
@@ -324,264 +433,24 @@ cleanup_user_failure() {
...
+(setup_user_admin_context && test_example_job_with_s3 && cleanup_user_success) || cleanup_user_failure
 
-(setup_user_admin_context && test_pyspark_in_pod && cleanup_user_success) || cleanup_user_failure_in_pod
+(setup_user_admin_context && test_example_job_with_azure_abfss && cleanup_user_success) || cleanup_user_failure
 
-(setup_user_admin_context && test_spark_sql_in_pod && cleanup_user_success) || cleanup_user_failure_in_pod
-
-(setup_user_admin_context && test_pyspark_s3_in_pod && cleanup_user_success) || cleanup_user_failure_in_pod
-
-teardown_test_pod
```
deusebio, line 587: **question** why did we remove the teardown?

deusebio (approved):
I have a couple of small points, but otherwise the code looks good! So just fix the niptick and then feel free to go ahead and merge it directly! Thanks!

tests/integration/ie-tests.sh
theoctober19th: The pod tests have been removed here (since they are almost duplicates of the tests done in charmed-spark-rock) and since we don't test on pod, the pod deletion and creation logic has been removed as well.

PR #112, opened 2025-05-09
[DPE-7131] Add beeline client to snap

Batalex:
Working
- Authenticated connection
- History file
- TLS

Testing locally:

- Clone the branch
- Pack the snap with `snapcraft pack`
- Install the locally packed charm with `sudo snap install ./spark-client_3.5.1_amd64.snap --dangerous`
- Run the client against a deployment with `spark-client.beeline -u "jdbc:hive2://10.1.111.103:10009/" -n admin -p password`. Please note that we need the unit IP and not the hostname provided by core dns

Using TLS:
- Get the CA from `self-signed-certificates` as usual (using the action, or through `data-integrator`) in a `ca.pem` file
- Import the .pem file into a truststore in the snap using `spark-client.import-certificate ca ./ca.pem`
- Use `spark-client.beeline -u "jdbc:hive2://10.1.111.103:10009/;ssl=true" -n user -p password`

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -134,12 +134,21 @@ parts:
       # Create the target directory and extract the tarball for Kyuubi
       KYUUBI_DIR=$CRAFT_PART_INSTALL/opt/kyuubi
       mkdir -p $KYUUBI_DIR/bin
+      mkdir -p $KYUUBI_DIR/logs
+      mkdir -p $KYUUBI_DIR/work
+      mkdir -p $KYUUBI_DIR/pid
+      mkdir -p $KYUUBI_DIR/externals/kyuubi-download/target
       mkdir kyuubi
       tar -xzvf $KYUUBI_TARFILE -C kyuubi/ --strip-components=1
 
+      # Resolve beeline jars symlinks, save ~50MB compared to copying kyuubi jars folder
+      for f in $(find kyuubi/beeline-jars -maxdepth 1 -type l)
+      do
+          cp --remove-destination $(readlink -e $f) $f
+      done
+
       # Stage beeline files
       cp -r kyuubi/beeline-jars $KYUUBI_DIR
-      cp -r kyuubi/jars $KYUUBI_DIR
       cp kyuubi/bin/kyuubi-beeline $KYUUBI_DIR/bin
       cp kyuubi/bin/load-kyuubi-env.sh $KYUUBI_DIR/bin
       chmod -R +x $KYUUBI_DIR/bin
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,6 +1,6 @@
 name: spark-client
 base: core22
-version: "3.5.1"
+version: "3.5.4"
 summary: Client side scripts to submit Spark jobs to a cluster.
 description: |
   The spark-client snap includes the scripts spark-submit, spark-shell, pyspark and other tools for managing Apache Spark jobs.
@@ -89,7 +89,7 @@ apps:
       - dot-kube-config
 
   beeline:
-    command: /opt/kyuubi/bin/kyuubi-beeline
+    command: /opt/spark/bin/beeline
     environment:
       _JAVA_OPTIONS: "-Duser.home=$SNAP_USER_DATA"
     plugs:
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -102,66 +109,102 @@ parts:
...
+      rm -vf usr/lib/jvm/java-17-openjdk-*/lib/security/blacklisted.certs
+      # move base certs data
+      mv etc/ssl/certs/java/cacerts etc/cacerts
+
+  kyuubi-beeline:
+    plugin: nil
+    build-packages:
+      - wget
+    override-build: |
+      KYUUBI_ARTIFACT="https://dlcdn.apache.org/kyuubi/kyuubi-1.10.1/apache-kyuubi-1.10.1-bin.tgz"
```
theoctober19th: **question**: IIRC, the `beeline` util is there in Spark default distribution itself. Can't we simply use that binary and not download Kyuubi artifact? I'm not sure how recent `/opt/spark/bin/beeline` is though -- however, I would still incline towards simply exposing this binary than having to add Kyuubi jars and artifacts, etc.

Batalex: Apparently, there are some [minor differences](https://kyuubi.readthedocs.io/en/master/client/cli/kyuubi_beeline.html#what-is-kyuubi-beeline) between hive beeline (bundled in spark) and kyuubi's beeline, so I don't think we should use spark's. 
Kyuubi's beeline adds about 70MB of jars, which is not negligible, of course. But AWS sdk by itself represents 300+MB in one jar so in the end I believe this is a fair addition. That's up to debate, of course.

One interesting thing is that there is only one common jar between the two folders from a quick visual inspection: `log4j-slf4j2-impl-2.20.0.jar`.
So there is not much we can do to reduce duplication right now

deusebio: **TLDR** I believe I agree with Bikalpa. 

I had a closer look at comparing the two CLIs, and they very very similar. The difference between the two are the following arguments:

```
  -c <named url>                  The named JDBC URL to connect to,
                                  which should be present in beeline-site.xml
                                  as the value of beeline.hs2.jdbc.url.<namedUrl>.

  --conf property=value           Alias of --hiveconf.
  --escapeCRLF=[true|false]       Show carriage return and line feeds in query results as escaped \r and \n.
  --outputformat=<format mode>    Format mode for result display.
                                  The available options ars [table|vertical|csv2|tsv2|dsv|csv|tsv|**json|jsonfile**].
                                  Note that csv, and tsv are deprecated, use csv2, tsv2 instead.

  --delimiter=DELIMITER           Set the query delimiter; multi-char delimiters are allowed, but quotation
                                  marks, slashes, and -- are not allowed (default: ;).

  --convertBinaryArrayToString=[true|false]
                                  Display binary column data as string or as byte array.

  --python-mode                   Execute python code/script.
  -h, --help                      Display this message.
``` 

So basically they are:
1. named aliases for the URL (the `-c` argument)
2. extended output formatting (`escapeCRLF`, `outputformat`, `delimiter`, `convertBinaryArrayToString`)
3. `python-mode`

The only rather interesting one in my point of view may be the `python-mode`, but maybe not worth adding all the kyuubi  solution (even though it is 70MB) IMHO.

snap/snapcraft.yaml
```diff
@@ -102,66 +109,102 @@ parts:
...
+      # Download and verify Kyuubi
+      wget -O $KYUUBI_TARFILE $KYUUBI_ARTIFACT
+      echo "$KYUUBI_512_CHECKSUM  $KYUUBI_TARFILE" | sha512sum -c -
+
+      # Create the target directories and extract the tarball for Kyuubi
+      KYUUBI_DIR=$CRAFT_PART_INSTALL/opt/kyuubi
+      mkdir -p $KYUUBI_DIR/bin
+      mkdir -p $KYUUBI_DIR/logs
+      mkdir -p $KYUUBI_DIR/work
+      mkdir -p $KYUUBI_DIR/pid
```
deusebio: **todo** I believe we should do a bit of organization of the folders here. Only `bin` and `jars` (and possibly `externals` but this is empty right now) should be in the squashfs system. The writable parts should be either in SNAP_COMMON or SNAP_DATA depending on their scope:
* logs and data into SNAP_COMMON
* configuration file into SNAP_DATA

there is [this spec](https://docs.google.com/document/d/1MyT4Wt_-vgPwhwEjC-oeHRdLYrZWaQzFGTwclT7hD1U/edit?tab=t.0) outlining a bit better the structure of snap packages

deusebio (changes requested):
After some thoughts, I believe I agree with Bikalpa, that to start with, I would probably only expose the beeline functionalities from Spark/Hive, rather than pulling the Kyuubi solution. Also packaging kyuubi as well falsely suggest to the user that he could use the snap to run the kyuubi deamon, which however this snap is not configured for. 

Also I have a comment about the structure of snap. But if we remove kyuubi, that may not be needed anymore though

snap/snapcraft.yaml
Batalex: I am happy to report that spark's beeline works just as well! You can give it a try using the same steps as before using 306ee9c171f49a8b9abcb919aae5e5b89242a0d6

snap/snapcraft.yaml
Batalex: The previous `beeline` entry point would block the execution should one of the folders mentioned here be missing, so one solution could have been to rewrite the `kyuubi-beeline.sh` script.
This is no longer an issue since 306ee9c171f49a8b9abcb919aae5e5b89242a0d6 reverts back to using the bundled beeline

deusebio (approved):
Looks good!!!

theoctober19th (approved):

welpaolo (approved):
LGTM!

PR #115, opened 2025-07-11
[DPE-7484] backport for artifact consolidation

deusebio:
This PR currently backports some of the latest features that landed on 3.5/edge branch.


snap/snapcraft.yaml
```diff
@@ -32,64 +36,71 @@ apps:
...
-        - home
-        - dot-kube-config
+      - network
+      - network-bind
+      - home
+      - dot-kube-config
 
-parts:
+  beeline:
+    command: /opt/spark/bin/beeline
```
welpaolo: Probably we should also add the PR that @Batalex just opened

welpaolo (approved):
LGTM. I left a comment.

snap/snapcraft.yaml
deusebio: yes, sure! Once that lands, I will add it here with the PR reference

deusebio:
backported changes from #116 and #114

.github/workflows/on_spark_update_available.yaml
```diff
@@ -1,7 +1,7 @@
 name: Detect upstream Spark updates
 
 env:
-  BASE_BRANCH: 3.4/edge
+  BASE_BRANCH: 3.5/edge
```
Batalex, line 4: **todo**: We should keep 3.4 on this branch, right?

Batalex (changes requested):

.github/workflows/on_spark_update_available.yaml
deusebio: I'm not sure. This is a scheduled process, so it is not a process that should run on both branches, and I thought that it would make more sense to run this with the latest one, ideally this should be for main TBH, but we don't have that here, so I fell back on the higher major.minor branch. 

But overall, I don't think this scheduled workflows is very valuable at the moment, also because if we start to pick the new versions, that would be 4/, for which we should have a new branch. 

All in all, probably the best way forward would be to remove this Github action all together, but I didn't want to do this in a backporting PR, so I would honestly keen the current PR as-is

Batalex (approved):

deusebio:
Superseded by #118

PR #116, opened 2025-07-11
fix: Remove leading slash

Batalex:
This PR removes the leading slash in the beeline command definition.
This was working locally, but leading to an error during the upload


deusebio (approved):
LGTM

welpaolo (approved):

PR #117, opened 2025-07-18
[DPE-6835] Housekeeping

Batalex:
There was not much left to do from the original ticket since the build job has been recently updated.


welpaolo (approved):

deusebio (approved):
Looks good!

PR #132, opened 2025-10-14
[DPE-8545] Update to Spark 4

welpaolo:
This PR added support for Spark 4.
This PR includes the restructure of tests dividing tls/non tls tests and use microceph instead of Minio.


.github/workflows/ci.yaml
```diff
@@ -60,3 +82,29 @@ jobs:
           AZURE_STORAGE_KEY: ${{ secrets.AZURE_STORAGE_KEY }}
         run: |
           sg microk8s -c "make integration-tests"
+      
+      - name: Setup microceph with SSL
```
deusebio, line 86: **question,out-of-scope** I remember that Pedro created an action to setup microceph. I'm wondering that maybe it would be good to give some love to (and use) the action such that we don't have to copy and paste this across ALL repos. 

This seems something very general we will need to do. It can be of course addressed in a different ticket that is about "consolidating microceph setup into re-usable action", possibily within the DP workflows

tests/integration/ie-tests.sh
```diff
@@ -43,15 +43,16 @@ setup_s3_properties(){
...
     --conf spark.hadoop.fs.s3a.connection.ssl.enabled=false \
     --conf spark.hadoop.fs.s3a.path.style.access=true \
+    --conf spark.hadoop.fs.s3a.fast.upload=true \
     --conf spark.hadoop.fs.s3a.endpoint=$(get_s3_endpoint) \
     --conf spark.hadoop.fs.s3a.access.key=$(get_s3_access_key) \
     --conf spark.hadoop.fs.s3a.secret.key=$(get_s3_secret_key) \
     --conf spark.sql.warehouse.dir=s3a://$S3_BUCKET/warehouse \
     --conf spark.sql.catalog.local.warehouse=s3a://$S3_BUCKET/warehouse \
-    --conf spark.hadoop.hive.metastore.warehouse.dir=s3a://$S3_BUCKET/hwarehouse  
+    --conf spark.hadoop.hive.metastore.warehouse.dir=s3a://$S3_BUCKET/warehouse
```
deusebio, line 55: **praise** what a catch!

Makefile
```diff
@@ -119,6 +119,18 @@ endif
...
 
+# Recipe for running integration tests with tls.
+integration-tests-tls: $(k8s_tag) $(aws_tag) $(azure_tag)
+ifndef SNAP_EXISTS
+	@echo "Installing snap first"
+	make install
+endif
+	@export AZURE_STORAGE_ACCOUNT=$(AZURE_STORAGE_ACCOUNT) \
+				AZURE_STORAGE_KEY=$(AZURE_STORAGE_KEY) \
+	&& sg microk8s tests/integration/ie-tests-tls.sh
```
deusebio: **question** `ie-tests-tls` seems to be connected only to S3, not to AZURE. Why are those env variable required?

tests/integration/ie-tests-tls.sh
```diff
@@ -0,0 +1,202 @@
...
+
+test_spark_submit_custom_certificate() {
+  run_spark_submit_custom_certificate
+}
+
+run_spark_submit_custom_certificate(){
+  
+  KUBE_CONFIG=/home/${USER}/.kube/config
+
+  # selete service account if it exist
```
deusebio:
```suggestion
  # delete service account if it exist
```

tests/integration/ie-tests-tls.sh
```diff
@@ -0,0 +1,202 @@
+#!/bin/bash
```
deusebio, line 1: **praise** REALLY GREAT TEST!!!

deusebio (commented):
Codewise, I only have very very minor comment and/or questions. Actually, I believe you have done an amazing job at creating  new test/check that can be very very good for spark 3.x as well. We have not been testing against a SSL/microceph with encryption, I believe it would be good to do so, regardless of the Spark 4 upgrades.

So the only major comment I have is that the current PR - as is - feels a bit overloading the scope with two independent things: 

1. Enabling TLS checks
2. Upgrading to Spark4

I understand that 1 is needed for 2, but I would honestly like to split the two, such that 1. can also be backported easily to the Spark 3 track (or probably landed there in the first place if that one is our development branch), and then a minimal Spark4 update PR (to update artifacts versions and AWS classes to be used)

.github/workflows/on_spark_update_available.yaml
```diff
@@ -1,7 +1,7 @@
 name: Detect upstream Spark updates
 
 env:
-  BASE_BRANCH: 3.5/edge
+  BASE_BRANCH: 4.0/edge
```
Batalex: **question**: Do we still need this file? We haven't had a run for a very long time, the action panel is empty + we use our built-from-source artifacts

Batalex (commented):
Changes LGTM; seems like you had to jump through quite a bunch of hoops to get TLS working. I'll refrain from approving right now, should we split this work into multiple PRs

.github/workflows/on_spark_update_available.yaml
deusebio: Uhm, maybe not. We can probably remove this IMHO

.github/workflows/ci.yaml
welpaolo: I created a [ticket](https://warthogs.atlassian.net/browse/DPE-8635) for this purpose.

PR #138, opened 2026-01-19
[PRA-144] Add arm64 support

welpaolo:
This PR add support for ARM64 architecture.

### DIFF
--- a/snap/local/etc/spark8t/bin/import-certificate.sh
+++ b/snap/local/etc/spark8t/bin/import-certificate.sh
@@ -7,7 +7,7 @@ if [[ $alias && $path ]]
 then
     echo "Alias $1"
     echo "Certificate path $2"
-    ${SNAP}/usr/lib/jvm/java-17-openjdk-amd64/bin/keytool -import -v -alias "$1" -file "$2"  -storepass changeit -noprompt -keystore ${SNAP_DATA}/etc/ssl/certs/java/cacerts
+    ${JAVA_HOME}/bin/keytool -import -v -alias "$1" -file "$2"  -storepass changeit -noprompt -keystore ${SNAP_DATA}/etc/ssl/certs/java/cacerts
     echo "Certificate imported!"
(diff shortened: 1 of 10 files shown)


snap/local/etc/spark8t/bin/import-certificate.sh
```diff
@@ -7,7 +7,7 @@ if [[ $alias && $path ]]
 then
     echo "Alias $1"
     echo "Certificate path $2"
-    ${SNAP}/usr/lib/jvm/java-17-openjdk-amd64/bin/keytool -import -v -alias "$1" -file "$2"  -storepass changeit -noprompt -keystore ${SNAP_DATA}/etc/ssl/certs/java/cacerts
+    ${JAVA_HOME}/bin/keytool -import -v -alias "$1" -file "$2"  -storepass changeit -noprompt -keystore ${SNAP_DATA}/etc/ssl/certs/java/cacerts
```
Batalex, line 10: **praise**: oh nice, we should have used that from the get go

tests/integration/setup-azure-cli.sh
```diff
@@ -1,7 +1,18 @@
 #!/bin/bash
 
-echo "Installing Azure CLI..."
-sudo snap install azcli
+if ! command -v az >/dev/null 2>&1
```
Batalex, line 3: **thought**: This is familiar :nerd_face:

Batalex (changes requested):
LGTM, but we should probably upgrade the image in `snap/local/etc/spark8t/spark-defaults.conf` to point to the newer manifest for both arm and amd, otherwise, the arm snap would not work by default on an arm machine without overriding the spark property

welpaolo:
Thanks for the catch Alex!

Batalex (approved):
Perfect, thanks!

theoctober19th (approved):
LGTM, thanks !

PR #149, opened 2026-03-11
[MISC] Reduce the K8s matrix validation and disable fail-fast feature

deusebio:
I noticed that when one of the workflows fail becuaes of any reasons, then all the other runs are cancelled. This seems a bit too strict, so I added the `fail-fast` option, and also reduced the K8s versions to just making it a bit more likely to happen

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,6 +20,7 @@ jobs:
       - build
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


welpaolo (approved):

Batalex (approved):
LGTM, sticking with even versions (not sure if all of them are LTS though) is probably good enough

PR #153, opened 2026-04-01
[PRA-42] Add support for Spark 4

theoctober19th:
This PR aims to add support for Spark 4 in Spark Client Snap.


.github/workflows/on_spark_update_available.yaml
Batalex, line 1: **suggestion**: We have not used this in a very long time, how about we drop it? This will most likely be replaced by a renovate configuration + we don't package release from upstream

SPARK_VERSION
```diff
@@ -1 +1 @@
-3.5.7-ubuntu2-20260212173650
\ No newline at end of file
+4.0.1-ubuntu7-20260126132057
```
Batalex, line 1: **suggestion**: We have spark-4.0.1-ubuntu8 available

theoctober19th: I thought keeping the same version in the rock image used in the default config + the tests, as the latest rock release was 4.0.1-ubuntu7. WDYT?

.github/workflows/on_spark_update_available.yaml
theoctober19th: That's a nice suggestion! I believe this is better done in a separate PR since other branches have this file too, and then backport into 4.0 later for clear history. I'll open the PR and link here.

theoctober19th: https://github.com/canonical/spark-client-snap/pull/154

SPARK_VERSION
Batalex: Well, let's keep this as-is, Paolo will be able to use this opportunity to test the renovate automation then.

Batalex (approved):

welpaolo (approved):

PR #155, opened 2026-04-02
[MISC] Upgrade data-platform-workflows to 48.0.2

theoctober19th:
Need this update to unblock the CI, because the release is now bottlenecked by 15 minutes timeout that was there in the older workflow..

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,6 +5,7 @@ on:
     branches:
       - 3.4/*
       - 3.5/*
+      - 4.0/*
 
 jobs:
   release_checks:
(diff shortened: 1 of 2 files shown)


welpaolo (approved):

.github/workflows/release.yaml
Batalex: **suggestion**: We might as well remove the risk everywhere since this is no longer used

theoctober19th: Fixed in [739462e](https://github.com/canonical/spark-client-snap/pull/155/commits/739462eeb50b443c709bdd58657c3a4aec3272aa).

Batalex (approved):

